PR #1, opened 2018-11-19
Support for loading the environment from the specified endpoint

tombuildsstuff:
Moving this up from [the Azure Stack Provider](https://github.com/terraform-providers/terraform-provider-azurestack/blob/master/azurestack/config.go#L145-L148) and adding some validation logic so that we can reuse it in the AzureRM Backend in the future

### DIFF
--- a/authentication/environment.go
+++ b/authentication/environment.go
@@ -25,6 +25,22 @@ func DetermineEnvironment(name string) (*azure.Environment, error) {
 	return &env, nil
 }
 
+// LoadEnvironmentFromUrl attempts to load the specified environment from the endpoint.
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
Travis was added after this PR so it won't build for it.. but `make` passes locally:

```
 $ make
go vet ./...

go test -race ./...
ok  	github.com/hashicorp/go-azure-helpers/authentication	2.452s
?   	github.com/hashicorp/go-azure-helpers/resourceproviders	[no test files]
ok  	github.com/hashicorp/go-azure-helpers/response	(cached)
```

PR #8, opened 2018-11-22
Splitting out the SAS Token Generation

tombuildsstuff:
Splitting this out from the AzureRM Repository, since it's needed in the Azure Backend

### DIFF
--- a/storage/sas_token.go
+++ b/storage/sas_token.go
@@ -1,0 +1,98 @@
+package storage
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+const (
+	connStringAccountKeyKey  = "AccountKey"
+	connStringAccountNameKey = "AccountName"
+)
+
+// ComputeSASToken computes the SAS Token for a Storage Account based on the
+// access key & given permissions
+func ComputeSASToken(accountName string,
+	accountKey string,
+	permissions string,
+	services string,
+	resourceTypes string,
+	start string,
+	expiry string,
+	signedProtocol string,
+	signedIp string, // nolint: unparam
+	signedVersion string, // nolint: unparam
+) (string, error) {
+
+	// UTF-8 by default...
+	stringToSign := accountName + "\n"
+	stringToSign += permissions + "\n"
+	stringToSign += services + "\n"
+	stringToSign += resourceTypes + "\n"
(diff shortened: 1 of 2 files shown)


storage/sas_token.go
```diff
@@ -0,0 +1,98 @@
...
+	// so its safe to include here for reference to understand the format.
+	// DefaultEndpointsProtocol=https;AccountName=azurermtestsa0;AccountKey=2vJrjEyL4re2nxCEg590wJUUC7PiqqrDHjAN5RU304FNUQieiEwS2bfp83O0v28iSfWjvYhkGmjYQAdd9x+6nw==;EndpointSuffix=core.windows.net
+	validKeys := map[string]bool{"DefaultEndpointsProtocol": true, "BlobEndpoint": true,
+		"AccountName": true, "AccountKey": true, "EndpointSuffix": true}
+	// The k-v pairs are separated with semi-colons
+	tokens := strings.Split(connString, ";")
+
+	kvp := make(map[string]string)
+
+	for _, atoken := range tokens {
```
alexsomesan: Some linters flag this form of `for` suggesting the `_` can be omitted.
Feel free to keep it like this if this is the convention throughout the codebase.

storage/sas_token_test.go
```diff
@@ -0,0 +1,98 @@
...
+		expectedAccountKey  string
+	}{
+		{
+			"DefaultEndpointsProtocol=https;AccountName=azurermtestsa0;AccountKey=2vJrjEyL4re2nxCEg590wJUUC7PiqqrDHjAN5RU304FNUQieiEwS2bfp83O0v28iSfWjvYhkGmjYQAdd9x+6nw==;EndpointSuffix=core.windows.net",
+			"azurermtestsa0",
+			"2vJrjEyL4re2nxCEg590wJUUC7PiqqrDHjAN5RU304FNUQieiEwS2bfp83O0v28iSfWjvYhkGmjYQAdd9x+6nw==",
+		},
+	}
+
+	for _, test := range testCases {
```
alexsomesan: Similar case of `for` with `_`

storage/sas_token.go
tombuildsstuff: as discussed irl - since this would be indexed as `int, string` this isn't possible

storage/sas_token_test.go
tombuildsstuff: (same here) this'd be int, value so we can't do this here afaik

PR #10, opened 2018-11-23
[WIP] New Auth Method: CloudShell

tombuildsstuff:
This PR adds a new authentication method specifically for CloudShell - which should allow us to deprecate the existing Azure CLI Parsing Auth.

This is mostly a duplication of the Azure CLI Parsing Auth at this point, with the notable exception that it assumes we're running in CloudShell.

Once this has been released we should be able to remove support for the Azure CLI Parsing Auth.

### DIFF
--- a/authentication/auth_method_cloudshell.go
+++ b/authentication/auth_method_cloudshell.go
@@ -1,0 +1,109 @@
+package authentication
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure/cli"
+	"github.com/hashicorp/go-multierror"
+)
+
+type cloudShellAuth struct {
+	profile *azureCLIProfile
+}
+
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
In retrospect this may not be necessary since it appears CloudShell is (somehow) forcing MSI ¯\_(ツ)_/¯

tombuildsstuff:
Confirmed: CloudShell's force-enabling MSI 🤷‍♂️ - so I guess we can rip this out entirely:

```
$ cat /usr/local/bin/terraform
#!/usr/bin/env bash

set -e

export ARM_SUBSCRIPTION_ID=`az account show --output=json | jq -r -M '.id'`
export ARM_TENANT_ID=`az account show --output=json | jq -r -M '.tenantId'`

export ARM_MSI_ENDPOINT=$MSI_ENDPOINT
if [ -z "$ARM_MSI_ENDPOINT" ]; then
  export ARM_USE_MSI=false
else
  export ARM_USE_MSI=true
fi
export PATH=/usr/local/terraform:$PATH

terraform "$@"
```

PR #21, opened 2018-12-12
adding support for user MSI authentication

### LINKED ISSUE #18: auth: support for multiple user assigned identities
As originally reported in https://github.com/terraform-providers/terraform-provider-azurerm/issues/2441 by @LaurentLesle it's possible for a Virtual Machine / Virtual Machine Scale Set to have multiple identities assigned.

When using Managed Service Identity for authentication at this time we only support authenticating using a single identity; we'd need to get [Azure/go-autorest](https://github.com/Azure/go-autorest) updated to be able to (optionally) specify the Object ID/Client ID ([as documented here](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http)) which should be used to obtain the credentials from the metadata endpoint.

Once that's available it should be possible for users to specify a custom Client ID which is passed to this function as a part of the Managed Service Identity auth method

LaurentLesle:
fixing the issue #18 

(fixes #18)


katbyte (approved):
Hi @LaurentLesle,

thank you for this PR! I hope you don't mind but i pushed a couple changes so we could get this merged today. mainly it needed the MSI auth tests updated with the new property. LGTL now! 👍

mbfrahry (approved):
LGTM!

katbyte:
@LaurentLesle,

Looks like you've not signed the CLA, any chance you could do that so we could get this merged? thanks!

LaurentLesle:
Hi @katbyte I have just signed the CLA. Anything else I have to do? Looks like the status is still pending

katbyte:
I'll force a recheck and see if that helps

LaurentLesle:
@katbyte looks like there is a miss-match between my git profile name "Laurent Lesle" used at the time of the commit and this is not reconciled automatically. I suggest you take the code and submit a new PR as it will be quicker to fix that me trying to fix the CLA.

jontreynes:
Should this have resolved the terraform issues using MSI?
I am using Terraform 0.12.28 with the azurerm provider

with this
```
terraform init -upgrade -input=false \
-backend-config="./Config/backend.$environment.tfvars" \
-backend-config="subscription_id=$subscriptionId" \
-backend-config="tenant_id=$tenantId" \
-backend-config="use_msi=true"
```
where /Config/backend.$environment.tfvars holds the values for 

```
storage_account_name = "PLACEHOLDER"
container_name       = "PLACEHOLDER"
key                  = "PLACEHOLDER"
```

However, I'm receiving an error
image

I've followed the documentation here
https://www.terraform.io/docs/backends/types/azurerm.html#subscription_id

CC: @tombuildsstuff , @LaurentLesle , @katbyte

PR #25, opened 2019-04-23
passing the sender to the request method

tombuildsstuff:
This allows the same sender to be sent for both the Auth requests and Client requests; which should ensure the proxy configuration is the same

### DIFF
--- a/authentication/auth_method_client_cert.go
+++ b/authentication/auth_method_client_cert.go
@@ -41,7 +41,7 @@ func (a servicePrincipalClientCertificateAuth) name() string {
 	return "Service Principal / Client Certificate"
 }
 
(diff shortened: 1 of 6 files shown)


authentication/auth_method_client_cert.go
```diff
@@ -58,6 +58,8 @@ func (a servicePrincipalClientCertificateAuth) getAuthorizationToken(oauthConfig
 		return nil, err
 	}
 
+	spt.SetSender(sender)
```
paultyng: is it valid to pass `nil`? should you avoid setting it if so?

tombuildsstuff: the SDK handles that internally and uses it's own client, as we do today - so this should be fine

PR #36, opened 2019-08-19
Auth: exposing the Primary and Multi Tenant Authorizers

tombuildsstuff:
This PR does a couple of things necessary to be able to use the Key Vault and Storage Authorizers within the Azure Providers:

1. Exposes both the Primary and Multi Tenant authorizers to be able to return an Authorizer for the Primary Tenant
2. .. via the new method `BearerAuthorizerCallback` which can be used to obtain a BearerAuthorizerCallback for the Primary Tenant - which in turn can be used for the Key Vault and Storage Authorizers which require using the Primary Tenant

There's a couple of renames in here too:

1. Renaming the struct `MultiOAuth` struct to `AuthConfig`
2. Renaming the func `GetMultiOAuthConfig` to `BuildOAuthConfig`, since it's building the OAuth Configuration

Tests pass:

```
$ make test
go mod download
warning: pattern "all" matched no module dependencies
go vet ./...
go test -race ./...
ok  	github.com/hashicorp/go-azure-helpers/authentication	1.024s
?   	github.com/hashicorp/go-azure-helpers/resourceproviders	[no test files]
ok  	github.com/hashicorp/go-azure-helpers/response	1.023s
?   	github.com/hashicorp/go-azure-helpers/sender	[no test files]
ok  	github.com/hashicorp/go-azure-helpers/storage	1.016s
```

### DIFF
--- a/authentication/config.go
+++ b/authentication/config.go
@@ -25,7 +25,7 @@ type Config struct {
 	authMethod authMethod
 }
 
-type AuthConfig struct {
+type OAuthConfig struct {
 	OAuth            *adal.OAuthConfig
 	MultiTenantOauth *adal.MultiTenantOAuthConfig
 }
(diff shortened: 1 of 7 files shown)


authentication/config.go
```diff
@@ -25,7 +25,7 @@ type Config struct {
 	authMethod authMethod
 }
 
-type MultiOAuth struct {
+type AuthConfig struct {
```
radeksimko: Nitpick - should't this be called `OAuthConfig`?

katbyte: I agree, going to change that now

katbyte: On second thought i think `MultiOAuthConfig` is more appropriate to differentiate it, it handles multiple oauth config types

tombuildsstuff: 👍 I'll switch this to `OAuthConfig` - in renaming this I was hoping to avoid reusing the term `Multi` here not to overload it, given that's also being used for `Multi Tenant Auth`

PR #41, opened 2019-10-22
add possibility to return connection string outputs from storage sas

aczelandi:
Required for [this](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1868) issue.

Adds helper methods to return the sas connection string and the connection strings specific to one of the account type: blob, file, queue, table.

I am preparing another PR in [terraform-provider-azurerm](https://github.com/terraform-providers/terraform-provider-azurerm) to integrate this change and fix the above issue.

### DIFF
--- a/storage/sas_token.go
+++ b/storage/sas_token.go
@@ -72,14 +72,12 @@ func ComputeAccountSASToken(accountName string,
 // ComputeAccountSASConnectionString computes the composed SAS Connection String for a Storage Account based on the
 // sas token
 func ComputeAccountSASConnectionString(accountName string, sasToken string) string {
-	connectionString := ""
-	connectionString += fmt.Sprintf("BlobEndpoint=https://%s.blob.core.windows.net/;", accountName)
-	connectionString += fmt.Sprintf("FileEndpoint=https://%s.file.core.windows.net/;", accountName)
-	connectionString += fmt.Sprintf("QueueEndpoint=https://%s.queue.core.windows.net/;", accountName)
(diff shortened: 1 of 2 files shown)


storage/sas_token.go
```diff
@@ -68,6 +69,27 @@ func ComputeAccountSASToken(accountName string,
...
+		"BlobEndpoint=https://%[1]s.blob.core.windows.net/;"+
+			"FileEndpoint=https://%[1]s.file.core.windows.net/;"+
+			"QueueEndpoint=https://%[1]s.queue.core.windows.net/;"+
+			"TableEndpoint=https://%[1]s.table.core.windows.net/;"+
+			"SharedAccessSignature=%[2]s", accountName, sasToken[1:]) // need to cut the first character '?' from the sas token
+}
+
+// ComputeAccountSASConnectionUrlForType computes the SAS Connection String for a Storage Account based on the
+// sas token and the storage type
+func ComputeAccountSASConnectionUrlForType(accountName string, sasToken string, storageType string) (string, error) {
```
tombuildsstuff: in general this library follows a pattern of returning a nil-object when there's an error, so that any attempt to access a bad result will panic - could we make this:

```suggestion
func ComputeAccountSASConnectionUrlForType(accountName string, sasToken string, storageType string) (*string, error) {
```

and then return `nil, fmt.Errorf..` below?

aczelandi:
Thanks for the input @tombuildsstuff! I adjusted my changes according to your comments.

PR #46, opened 2019-11-18
report all errors in resource providers registration

### LINKED ISSUE #45: Resource provider RegisterForSubscription returns only last error
If RegisterForSubscription fails for multiple provider only the last error is returned:

https://github.com/hashicorp/go-azure-helpers/blob/6e061044db42917eeaecea091acb19e5d917122e/resourceproviders/registration.go#L40-L48

It should return all error and/or all providers that fail to register.

If someone have not enough permission to enable those provider it would be very helpful to have a error message with all resource provider that are unable to register.

Related: https://github.com/terraform-providers/terraform-provider-azurerm/issues/4799

StefanSchoof:
Collect all errors in resource provider registration.
My first go code. So expect go beginners mistakes.

Resolves #45

### DIFF
--- a/resourceproviders/registration.go
+++ b/resourceproviders/registration.go
@@ -55,7 +55,10 @@ func RegisterForSubscription(ctx context.Context, client resources.ProvidersClie
 
 	wg.Wait()
 
-	return fmt.Errorf("Cannnot register providers: %s. Errors were: %w", strings.Join(failedProviders, ", "), err)
+	if len(failedProviders) > 0 {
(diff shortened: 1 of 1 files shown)


StefanSchoof:
I tried this with go 1.13 (same as terraform-provider-azurerm). Should I make this 1.11 compatible or should the version bumped to 1.13?

StefanSchoof:
I just realized, this is not working for the non error case. Will try to fix it.

StefanSchoof:
It should now work also in the non error case and I also make it work with the old go version

tombuildsstuff (approved):
LGTM - thanks for this @StefanSchoof

PR #52, opened 2020-04-20
Load token/profile from CLI output (rather than file)

magodo:
## Motivation

As Az CLI is gonna deprecate accessTokens.json and azureProfile.json in the near future, we shall not loading them from file.

## Background

Az CLI comes with commands which has same effect as parsing files:

- `az account show [-s subid]`: returns a json object representing the
  specified subscription object (the default one if `subid` not
  specified)
- `az account get-access-token`: returns a json object representing the
  specified token object.

## Current Situation

Current implementation, when building for azure cli auth method, it loads the *azureProfile.json* file and parse it into a `profile` object. Then it will resolve the`subscriptionId` and populate the `clientId`,
`tenantId`, `environment` up to the `azureCLIProfile`.

Especially, when populating `clientId`, it will load *accessTokens.json* file and iterates all the token objects contained in it to find the "arm" token with correct `tenantId`, use its `clientId` to populate back.

## Changes

The fact is that there is no way to get either `clientID` or `refreshToken` from the output of `az account get-access-token`. On the other hand, the reason go-autorest depends on `clientID` and `refreshToken` is that they will be used to refresh the accessToken. While, go-autoreset will invoke the refresh token logic **only when** there is no custom refresh function set. In our case we are setting the custom refresh logic to call `az cli get-access-token` to refresh the new access token. This means, there is no need to set `clientID`/`refreshToken` in our case.
(go-autorest will check empty for `clientID` though, so I just pass in a dummy clientID to make the validation pass).

Another change is that rather than reading from *azureProfile.json* to get all the subscriptions, we can just read the one user specified in provider block/env var via `az account show -s [subid]` . Or loading the default one if user didnt explicitly specified that.


authentication/auth_method_azure_cli_token.go
```diff
@@ -22,28 +31,28 @@ type azureCliTokenAuth struct {
...
-			environment:    b.Environment,
 			subscriptionId: b.SubscriptionID,
 			tenantId:       b.TenantID,
+			// Setting a non-empty string as client ID to make adal.NewServicePrincipalTokenFromManualToken() happy
+			// (See: https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L473)
+			// The returned ServicePrincipalToken will only use that client ID when using the default logic to refresh
+			// the token, however, as we are using custom function to refresh token (i.e. using az cli) instead, it is
+			// not necessary to pass in a valid client ID.
+			// Actually there is no formal way to get either the client ID or refresh token from Az CLI.
+			clientId:       "cli",
```
tombuildsstuff, line 42: This'll cause issues in client configurations where a users trying to pass the Client ID in - as such we should probably hard-code the Client ID of the Azure CLI here instead - which clearly isn't ideal.

magodo: I feel for cli authentication, it makes less sense to allow user to specify anything other than subscription id. I'm not sure, but is there any use case that user need to specify a subscription ID together with tenant ID/client ID for authentication using az cli?

magodo: I understand your concern now, I updated the code to hardcoded that cli client id inline.

katbyte: I agree with tom we shouldn't change the behaviour here and use the CLI's client ID

magodo:
@tombuildsstuff Do we have any update on this PR?

LaurentLesle:
@tombuildsstuff I have noticed the command you are running is `az account show`. In the recent past I have been impacted by a global setting changing the default output format to a different value than json `az configure -o table`. Would be good to check that this setting does not impact the output of the az cli when called by terraform. If that's the case using `az account show -o json` will force the output to json

magodo:
@LaurentLesle Good catch! I'll submit a fix to explicitly setting the format.

authentication/auth_method_azure_cli_token.go
```diff
@@ -179,6 +182,23 @@ func obtainAuthorizationToken(endpoint string, subscriptionId string) (*cli.Toke
 	return &token, nil
 }
 
+// obtainSubscription return a subscription object of the specified subscriptionId.
+// If the subscriptionId is empty, it returns the default subscription.
+func obtainSubscription(subscriptionId string) (*cli.Subscription, error) {
+	var sub cli.Subscription
+	cmd := make([]string, 2, 4)
```
tombuildsstuff: this starts at 3 now, and we can remove the max (since it's invalid)

```suggestion
	cmd := make([]string, 0)
```

PR #53, opened 2020-05-07
Support for loading an Azure environment from a file

igal-s:


tombuildsstuff:
hey @igal-s 

Thanks for this PR - apologies for the delayed review here!

We've recently spent some time looking into this - and rather than pulling this information from a file on disk, we've instead opted to retrieve the environment information from the metadata endpoint (as implemented in #56) - which allows this to be dynamically configured/updated by the service.

As such - whilst I'd like to thank you for this contribution I'm going to close this PR for the moment since I believe that https://github.com/terraform-providers/terraform-provider-azurerm/issues/6816 should ultimately be fixed by #56.

Thanks!

PR #59, opened 2020-07-24
add check for when metadata url isn't passed in

mbfrahry:


authentication/environment.go
```diff
@@ -119,7 +123,7 @@ func AzureEnvironmentByNameFromEndpoint(ctx context.Context, endpoint string, en
 		}
 	}
 
-	return nil, fmt.Errorf("unable to find environment %q from endpoint %q", environmentName, endpoint)
+	return nil, fmt.Errorf("unable to locate metadata from environment %q and custom metadata host %q", environmentName, endpoint)
```
katbyte, line 126:
```suggestion
	return nil, fmt.Errorf("unable to locate metadata for environment %q from custom metadata host %q", environmentName, endpoint)
```

authentication/environment.go
```diff
@@ -103,6 +103,10 @@ func AzureEnvironmentByNameFromEndpoint(ctx context.Context, endpoint string, en
 		return &env, nil
 	}
 
+	if endpoint == "" {
+		return nil, fmt.Errorf("unable to locate metadata for environment %q from the built in ones `public`, `usgoverment`, `china` and no custom metadata host has been specified", environmentName)
```
katbyte, line 107:
```suggestion
		return nil, fmt.Errorf("unable to locate metadata for environment %q from the built in `public`, `usgoverment`, `china` and no custom metadata host has been specified", environmentName)
```

PR #60, opened 2020-07-25
Support configuring a Tenant ID and ignoring subscriptions

manicminer:
Support tenant-only configurations with or without associated subscription(s)

See testing matrix at https://docs.google.com/spreadsheets/d/1HXJrTLPloqsIaZNnv4ynhyy_9X5ibbW4BAG9zr0oCrE

### Changes

* Add a new Builder field `TenantOnly`.
* When true for CLI authentication, locates a suitable CLI 'account' by tenant ID instead of by subscription ID, and `SubscriptionID` is ignored / not required.
* Works with or without the `--allow-no-subscriptions` argument to `az login`.
* Will now support versions of Azure CLI that do or don't populate tenant-level accounts where a linked subscription is found.
* For all auth methods, permits validation without specifying `SubscriptionID`.
* Allows use of modules that don't operate on subscriptions (e.g. graphrbac).

### Related

- https://github.com/terraform-providers/terraform-provider-azuread/issues/294
- https://github.com/terraform-providers/terraform-provider-azuread/issues/314

### DIFF
--- a/authentication/auth_method_azure_cli_token.go
+++ b/authentication/auth_method_azure_cli_token.go
@@ -240,6 +240,10 @@ func obtainTenant(tenantId string) (*cli.Subscription, error) {
 				break
 			}
 		}
+
+		if acc.TenantID == "" {
+			return nil, fmt.Errorf("Tenant %q was not found", tenantId)
+		}
 	}
 
 	return &acc, nil
--- a/authentication/auth_method_azure_cli_token.go
+++ b/authentication/auth_method_azure_cli_token.go
@@ -15,12 +15,14 @@ import (
 )
 
 type azureCLIProfile struct {
-	subscription *cli.Subscription
+	// CLI "subscriptions" are really "accounts" that can represent either a subscription (with tenant) or _just_ a tenant
+	account *cli.Subscription
 
 	clientId       string
 	environment    string
 	subscriptionId string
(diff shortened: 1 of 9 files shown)


tombuildsstuff (approved):
LGTM 👍

authentication/auth_method_azure_cli_token.go
```diff
@@ -182,21 +196,53 @@ func obtainAuthorizationToken(endpoint string, subscriptionId string) (*cli.Toke
...
+			return nil, fmt.Errorf("Error parsing json result from the Azure CLI: %v", err)
+		}
+
+		for _, a := range accs {
+			if a.TenantID == tenantId {
+				acc = a
+				break
+			}
+		}
+	}
```
tombuildsstuff, line 243: edge-case but what if this isn't found - should we raise an error here?

tombuildsstuff (approved):
one minor comment but this otherwise LGTM 👍

authentication/auth_method_azure_cli_token.go
manicminer: Good call, not necessarily an edge case

manicminer: An error should now be raised when the specified tenantId is not found in az accounts

PR #65, opened 2020-09-07
cli auth result depends on exit code

magodo:
Do not simply return when stderr is not empty, since az cli might output warning to stderr. Instead, checking the returned `err` of `cmd.Wait()` is enough to tell whether the command exited with status code 0 or not.

This fixes: terraform-providers/terraform-provider-azurerm#8358


tombuildsstuff (approved):
LGTM 👍

PR #66, opened 2020-10-12
adding detailed error when az command failed

magodo:
The `err` only contains information like: "exit status: 1", which do not provide helpful information when trouble shooting.


tombuildsstuff (approved):
LGTM (I've pushed a commit to fix the build issue after the rebase) 👍

PR #68, opened 2020-10-28
Select the correct access token in a multi tenant scenario and subscription ID is not specified

manicminer:
Note
- The `-t` argument to `az account get-access-token` is "preview-only"
  but is included in recent releases of azure-cli by default
- Also includes a version check for Azure CLI, returning a detailed error if the available version is too old

Tested on a fresh azure-cli v2.14.0 installation on a new debian VM, with these scenarios: https://docs.google.com/spreadsheets/d/1HXJrTLPloqsIaZNnv4ynhyy_9X5ibbW4BAG9zr0oCrE/edit#gid=0

### DIFF
--- a/authentication/auth_method_azure_cli_token.go
+++ b/authentication/auth_method_azure_cli_token.go
@@ -191,8 +191,10 @@ func (a azureCliTokenAuth) validate() error {
 }
 
 func (a azureCliTokenAuth) checkAzVersion() error {
+	// Azure CLI v2.0.79 is the earliest version to have a `version` command
 	var minimumVersion string
 	if a.profile.tenantOnly {
+		// v2.0.81 introduced the `--tenant` option to the `account get-access-token` subcommand
 		minimumVersion = "2.0.81"
 	} else {
 		minimumVersion = "2.0.79"
--- a/authentication/auth_method_azure_cli_token.go
+++ b/authentication/auth_method_azure_cli_token.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure/cli"
 	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-version"
 )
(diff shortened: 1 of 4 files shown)


authentication/auth_method_azure_cli_token.go
```diff
@@ -172,6 +190,50 @@ func (a azureCliTokenAuth) validate() error {
...
+		return fmt.Errorf("Please ensure you have installed Azure CLI version %s or newer. Error parsing json result from the Azure CLI: %v.", minimumVersion, err)
+	}
+
+	if cliVersion.AzureCli == nil {
+		return fmt.Errorf("Could not detect Azure CLI version. Please ensure you have installed Azure CLI version %s or newer.", minimumVersion)
+	}
+
+	actual, err := version.NewVersion(*cliVersion.AzureCli)
+	if err != nil {
+		return fmt.Errorf("Could not parse detected Azure CLI version: %+v", err)
```
tombuildsstuff, line 220:
```suggestion
		return fmt.Errorf("Could not parse detected Azure CLI version %q: %+v", *cliVersion.AzureCli, err)
```

authentication/auth_method_azure_cli_token.go
```diff
@@ -172,6 +190,50 @@ func (a azureCliTokenAuth) validate() error {
...
+	}
+
+	supported, err := version.NewVersion(minimumVersion)
+	if err != nil {
+		return fmt.Errorf("Could not parse supported Azure CLI version: %+v", err)
+	}
+
+	if supported.LessThanOrEqual(actual) {
```
tombuildsstuff, line 228: maybe overkill, do we want to check it's less than 3.0 too?

tombuildsstuff (approved):
a few minor points but this otherwise LGTM 👍

authentication/auth_method_azure_cli_token.go
manicminer: Good call, have implemented :+1:

PR #69, opened 2021-01-08
update github.com/Azure/go-autorest/autorest/adal to v0.9.10

magodo:
Update the adal dependency to include latest fixes.
The main purpose is to include this Azure/go-autorest#594, which solves the problem that users can't use the managed identity to authenticate Terraform in Azure App Function environment, where the payload format is different than it in other resource (like VM).


tombuildsstuff:
hi @magodo 

Thanks for this PR.

Since Terraform intentionally doesn't support being run inside of an App Service and Function App at this time, whilst I'd like to thank you for this contribution I'm going to close this PR in favour of #71 - which upgrades these dependencies but explicitly opts out of supporting MSI when running inside App Service/Function Apps.

Thanks!

PR #71, opened 2021-02-25
msi: explicitly opting out of support for App Service and Function Apps

tombuildsstuff:
At this time we only support using MSI on a Virtual Machine and CloudShell - as such at this time we intentionally opt-out of supporting MSI inside App Service and Function Apps (available in the latest `Azure/go-autorest` update) due to known issues within Terraform Core when running in a memory constrained environment


jackofallops (approved):
LGTM 👍

PR #74, opened 2021-04-26
upstreaming: moving Resource ID parsing into this package

tombuildsstuff:
Ultimately this enables us to consolidate this into a single place - this also introduces a `NewLongRunningPollerFromResponse` for the Poller


jackofallops (approved):
LGTM 👍

PR #75, opened 2021-05-11
formatting: introducing a package for parsing dates in a given format

tombuildsstuff:
This allows us to contain this within a single place, since this is necessary in multiple downstream dependencies.


jackofallops (approved):
LGTM 👍

PR #79, opened 2021-06-07
Backport of id.Provider logic

aristosvo:
Original PR https://github.com/terraform-providers/terraform-provider-azurerm/pull/12076

PR #80, opened 2021-06-08
Fix: Assign `subscriptionID` with a value

aristosvo:
I was to hasty with https://github.com/hashicorp/go-azure-helpers/pull/79 it seems 😢


katbyte (approved):
LGTM 👍

PR #81, opened 2021-06-15
Update logging for possible artificial response error 

yupwei68:
When fixing track2 error https://github.com/Azure/azure-sdk-for-go/issues/14729 , we find that sometimes the response returns nil, but the error is not nil.


tombuildsstuff (approved):
LGTM - thanks for this @yupwei68

PR #83, opened 2021-11-01
Support for oauth v2 tokens, Microsoft Graph, and a fix for Azure CLI 2.30.0

manicminer:
Depends on: https://github.com/manicminer/hamilton/pull/123 (~needs rebase accordingly prior to merge~ *done*)
Related: https://github.com/hashicorp/terraform-provider-azurerm/pull/14058

- Fix for Azure CLI 2.30.0 when using multi-tenant auth

- Support for obtaining v2 oauth tokens (i.e. MSAL)
  - Uses Hamilton to supply a compatible authorizer for autorest
  - Breaking change: function signatures for `authentication.GetAuthorizationToken` and `authentication.BearerAuthorizerCallback` have changed to maintain compatibility with new `authentication.GetAuthorizationTokenV2` function

- Opt-in ability to query Microsoft Graph for authenticated principal information

- Remove tests/support for German national cloud


authentication/auth_method_client_secret.go
```diff
@@ -1,25 +1,33 @@
...
 
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/hashicorp/go-multierror"
+	"github.com/manicminer/hamilton/auth"
+	"github.com/manicminer/hamilton/environments"
 )
 
 type servicePrincipalClientSecretAuth struct {
+	ctx            context.Context
```
tombuildsstuff, line 16: rather than doing this could we add `ctx` to the method parameters below? as a general rule you shouldn't store the context in Go afaik

authentication/auth_method_client_secret_multi_tenant.go
```diff
@@ -19,8 +25,10 @@ type servicePrincipalClientSecretMultiTenantAuth struct {
 
 func (a servicePrincipalClientSecretMultiTenantAuth) build(b Builder) (authMethod, error) {
 	method := servicePrincipalClientSecretMultiTenantAuth{
+		ctx:                b.Context,
```
tombuildsstuff, line 28: (as above)

authentication/azure_sp_objectid.go
```diff
@@ -3,47 +3,144 @@ package authentication
...
 
 func buildServicePrincipalObjectIDFunc(c *Config) func(ctx context.Context) (string, error) {
 	return func(ctx context.Context) (string, error) {
-		env, err := AzureEnvironmentByNameFromEndpoint(ctx, c.MetadataHost, c.Environment)
-		if err != nil {
+		objectId, err := objectIdFromMsGraph(ctx, c)
+		if err == nil {
+			return objectId, nil
+		}
+		if !strings.HasPrefix(err.Error(), "access denied") {
```
tombuildsstuff, line 26: should this be:

```suggestion
		if !strings.Contains(strings.ToLower(err.Error()), "access denied") {
```

authentication/azure_sp_objectid.go
```diff
@@ -3,47 +3,144 @@ package authentication
...
 			return "", err
 		}
 
-		s := sender.BuildSender("GoAzureHelpers")
+		log.Printf("Unable to retrieve service principal object ID from Microsoft Graph: %+v", err)
+		return objectIdFromAadGraph(ctx, c)
```
tombuildsstuff, line 31: we should have the context on the `config` here to know which to use, so perhaps we don't need this fallback here?

authentication/azure_sp_objectid.go
```diff
@@ -3,47 +3,144 @@ package authentication
...
+	}
+
+	if listResult.Values() == nil || len(listResult.Values()) != 1 || listResult.Values()[0].ObjectID == nil {
+		return "", fmt.Errorf("unexpected Service Principal query result: %#v", listResult.Values())
+	}
+
+	return *listResult.Values()[0].ObjectID, nil
+}
+
+func objectIdFromMsGraph(ctx context.Context, c *Config) (string, error) {
```
tombuildsstuff, line 71: can we return a pointer here to ensure we don't use an invalid value accidentally (means we'll crash if we attempt to run without handling the error, which'll get caught at acctest time):

```suggestion
func objectIdFromMsGraph(ctx context.Context, c *Config) (*string, error) {
```

authentication/azure_sp_objectid.go
```diff
@@ -3,47 +3,144 @@ package authentication
...
+		log.Printf("Unable to retrieve service principal object ID from Microsoft Graph: %+v", err)
+		return objectIdFromAadGraph(ctx, c)
+	}
+}
 
-		oauthConfig, err := c.BuildOAuthConfig(env.ActiveDirectoryEndpoint)
-		if err != nil {
-			return "", err
-		}
+func objectIdFromAadGraph(ctx context.Context, c *Config) (string, error) {
```
tombuildsstuff, line 35: (same here)

authentication/azure_sp_objectid.go
```diff
@@ -3,47 +3,144 @@ package authentication
...
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
+	"github.com/manicminer/hamilton/auth"
+	"github.com/manicminer/hamilton/environments"
+	"github.com/manicminer/hamilton/msgraph"
+	"github.com/manicminer/hamilton/odata"
+
 	"github.com/hashicorp/go-azure-helpers/sender"
 )
 
 func buildServicePrincipalObjectIDFunc(c *Config) func(ctx context.Context) (string, error) {
```
tombuildsstuff, line 20: we should probably make this:

```suggestion
func buildServicePrincipalObjectIDFunc(c *Config) func(ctx context.Context) (*string, error) {
```

authentication/azure_sp_objectid.go
```diff
@@ -3,47 +3,144 @@ package authentication
...
+
+	msGraphAuth, err := c.GetAuthorizationTokenV2(sender.BuildSender("GoAzureHelpers"), oauthConfig, string(env.MsGraph.Endpoint))
+	if err != nil {
+		return "", err
+	}
+
+	authorizerWrapper, err := auth.NewAutorestAuthorizerWrapper(msGraphAuth)
+	if err != nil {
+		return "", err
+	}
```
tombuildsstuff, line 90: could we wrap these errors?

authentication/builder.go
```diff
@@ -13,6 +13,9 @@ var (
 // Builder supports all of the possible Authentication values and feature toggles
 // required to build a working Config for Authentication purposes.
 type Builder struct {
+	// Context for passing to auth methods that accept one
+	Context context.Context
```
tombuildsstuff, line 17: (as above) can we move this onto the methods rather than holding this here, since this'll be different (with a different cancellation token) each time

go.mod
```diff
@@ -2,16 +2,19 @@ module github.com/hashicorp/go-azure-helpers
...
 	github.com/hashicorp/go-version v1.3.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.8.0
-	golang.org/x/crypto v0.0.0-20210421170649-83a5a9bb288b
+	github.com/manicminer/hamilton v0.33.1-0.20211104093353-2694c30d3542
+	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a
 )
 
 go 1.16
+
+replace github.com/manicminer/hamilton => /Users/tom/go/src/github.com/manicminer/hamilton
```
tombuildsstuff, line 20: calling this out :)

tombuildsstuff (commented):
Left a few comments inline, but as discussed this mostly looks like it's going in the right direction to me 👍

.python-version
```diff
@@ -0,0 +1 @@
+3.10.0/envs/azure-cli-2.30.0
```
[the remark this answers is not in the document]
manicminer: Oops, this was meant to be gitignored :)

authentication/auth_method_azure_cli_token_multi_tenant.go
```diff
@@ -11,14 +11,15 @@ import (
...
 	profile                      *azureCLIProfileMultiTenant
 	servicePrincipalAuthDocsLink string
 }
 
 func (a azureCliTokenMultiTenantAuth) build(b Builder) (authMethod, error) {
 	auth := azureCliTokenMultiTenantAuth{
+		clientId: b.ClientID,
 		profile: &azureCLIProfileMultiTenant{
-			clientId:           b.ClientID,
 			environment:        b.Environment,
```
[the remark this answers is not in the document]
manicminer: You're right, I was in a multi-tenant testing rabbithole and this is a vestige that shouldn't be here

authentication/auth_method_azure_cli_token_multi_tenant.go
```diff
@@ -158,16 +159,16 @@ func (a azureCliTokenMultiTenantAuth) populateConfig(c *Config) error {
...
 		return fmt.Errorf("Azure CLI Profile is nil - this is an internal error and should be reported.")
 	}
 
-	if a.profile.clientId == "" {
-		err = multierror.Append(err, fmt.Errorf(errorMessageFmt, "Client ID"))
+	if a.clientId == "" {
+		err = multierror.Append(err, fmt.Errorf("a %s must be configured when authenticating using the Azure CLI with Auxiliary Tenant IDs.", "Client ID"))
 	}
 
+	errorMessageFmt := "A %s was not found in your Azure CLI Credentials.\n\nPlease login to the Azure CLI again via `az login`"
```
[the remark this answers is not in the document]
manicminer: Given that v1 and v2 tokens are interoperable for the foreseeable future, I think we should allow it to work with either and delegate that to the user? Else we'd end up hard erroring which would make it easier for the user to simply roll back to v1 tokens all round.

authentication/auth_method_client_secret.go
manicminer: Makes sense, will look at updating these methods accordingly

authentication/azure_sp_objectid.go
manicminer: I tried to separate the concerns of selecting a token version and choosing which API to call, given that both token versions work with both APIs and there are different prerequisites.

Having it gracefully fall back is the first pass - I'm hoping to find a neat way of exposing this downstream so that a deprecation notice can be raised. WDYT?

manicminer: I think I misunderstood what you meant - we could make this configurable instead of having auto fallback. Actually now I'm thinking this would be better since it can be explicitly opted into, and when it fails it's directly due to user configuration which feels more obvious/transparent.

manicminer:
@tombuildsstuff Thanks for the initial review! I believe this is ready for further review, and should be close to merge quality pending https://github.com/manicminer/hamilton/pull/123

authentication/auth_method_azure_cli_token_multi_tenant.go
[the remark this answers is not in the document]
tombuildsstuff: Seems reasonable for now 👍

authentication/auth_method_client_cert.go
```diff
@@ -68,6 +76,36 @@ func (a servicePrincipalClientCertificateAuth) getAuthorizationToken(sender auto
...
+	}
+
+	conf := auth.ClientCredentialsConfig{
+		Environment:        environment,
+		TenantID:           a.tenantId,
+		AuxiliaryTenantIDs: a.auxiliaryTenantIds,
+		ClientID:           a.clientId,
+		PrivateKey:         x509.MarshalPKCS1PrivateKey(rsaPrivateKey),
+		Certificate:        certificate.Raw,
+		Scopes:             []string{fmt.Sprintf("%s/.default", strings.TrimRight(endpoint, "/"))},
```
tombuildsstuff, line 97: not a blocker, more a thought - we should add a helper for `DefaultScopeForEndpoint` which does this

authentication/auth_method_msi.go
```diff
@@ -74,6 +75,11 @@ func (a managedServiceIdentityAuth) getAuthorizationToken(sender autorest.Sender
 	return auth, nil
 }
 
+func (a managedServiceIdentityAuth) getMSALToken(ctx context.Context, sender autorest.Sender, oauthConfig *OAuthConfig, endpoint string) (autorest.Authorizer, error) {
+	// v2 tokens not supported, so we'll pass through to the existing method for continuity
```
tombuildsstuff, line 79: perhaps clearer as:

```suggestion
	// auth tokens come directly from the MSI endpoint, so we'll pass through to the existing method for continuity
```

authentication/config.go
```diff
@@ -98,28 +101,33 @@ func (c Config) BuildOAuthConfig(activeDirectoryEndpoint string) (*OAuthConfig,
...
 
 		return cast, nil
 	})
 }
 
-// GetAuthorizationToken returns an authorization token for the authentication method defined in the Config
-func (c Config) GetAuthorizationToken(sender autorest.Sender, oauth *OAuthConfig, endpoint string) (autorest.Authorizer, error) {
-	return c.authMethod.getAuthorizationToken(sender, oauth, endpoint)
+// GetAuthorizationToken returns an autorest.Authorizer for the authentication method defined in the Config
+func (c Config) GetAuthorizationToken(ctx context.Context, sender autorest.Sender, oauth *OAuthConfig, endpoint string) (autorest.Authorizer, error) {
```
tombuildsstuff, line 126: can we make this:

```suggestion
// GetADALToken returns an autorest.Authorizer using a ADAL Token via the authentication method defined in the Config
func (c Config) GetADALToken(ctx context.Context, sender autorest.Sender, oauth *OAuthConfig, endpoint string) (autorest.Authorizer, error) {
```

authentication/config.go
```diff
@@ -98,28 +101,33 @@ func (c Config) BuildOAuthConfig(activeDirectoryEndpoint string) (*OAuthConfig,
...
+	return c.authMethod.getADALToken(ctx, sender, oauth, endpoint)
+}
+
+// GetAuthorizationTokenV2 returns an autorest.Authorizer sourced from hamilton/auth
+func (c Config) GetAuthorizationTokenV2(ctx context.Context, sender autorest.Sender, oauth *OAuthConfig, endpoint string) (autorest.Authorizer, error) {
```
tombuildsstuff, line 131: can we make this:

```suggestion
// GetMSALToken returns an autorest.Authorizer using a MSAL Token via the authentication method defined in the config
func (c Config) GetMSALToken(ctx context.Context, sender autorest.Sender, oauth *OAuthConfig, endpoint string) (autorest.Authorizer, error) {
```

(this is leaking the implementation details, which will change over time)

tombuildsstuff (approved):
A few minor comments but this is otherwise G2G 👍

PR #84, opened 2021-11-03
resourcemanager: introducing an updated Resource ID Parser based on Segments

tombuildsstuff:
This allows the parsing of Resource ID's which contain Scopes either at the start or the end - but not in the middle (since we've not seen any) - as such this should be fine.

This means that Resource ID's now need to expose a slice of expected segments, for example:

```go
func (id ConfigurationStoreId) ID() string {
	fmtString := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.AppConfiguration/configurationStores/%s"
	return fmt.Sprintf(fmtString, id.SubscriptionId, id.ResourceGroupName, id.ConfigStoreName)
}

func (id ConfigurationStoreId) Segments() []resourceids.Segment {
	return []resourceids.Segments{
		resourceids.StaticSegment("subscriptions", "subscriptions", "exampleValue"),
		resourceids.SubscriptionIdSegment("subscriptionId", "exampleValue"),
		resourceids.StaticSegment("resourceGroups", "resourceGroups", "exampleValue"),
		resourceids.ResourceGroupSegment("resourceGroupName", "exampleValue"),
		resourceids.StaticSegment("providers", "providers", "exampleValue"),
		resourceids.ResourceProviderSegment("microsoftAppConfiguration", "Microsoft.AppConfiguration", "exampleValue"),
		resourceids.StaticSegment("configurationStores", "configurationStores", "exampleValue"),
		resourceids.UserSpecifiedSegment("configStoreName", "exampleValue"),
	}
}
```

This PR's based on the work done by @koikonom but instead uses a regex to find any Prefix Scope, then parses the segments we expect (based on the `Segments()` function above) - and finally raises an error if there's any unexpected values present / uses any remainder as the Suffix Scope if applicable.

Notably the `insensitive` version of this parser will case-correct any static segments to match what we're expecting, which whilst isn't perfect - should help normalise some of the API casing issues as Terraform Core requires these are byte-for-byte compatible.

### DIFF
--- a/resourcemanager/resourceids/interface.go
+++ b/resourcemanager/resourceids/interface.go
@@ -1,11 +1,125 @@
 package resourceids
 
-// Id defines a type for a ResourceId of some kind
-type Id interface {
+type ResourceId interface {
 	// ID returns the fully formatted ID for this Resource ID
 	ID() string
 
 	// String returns a friendly description of the components of this Resource ID
 	// which is suitable for use in error messages (for example 'MyThing %q / Resource Group %q')
 	String() string
+
+	// Segments returns an ordered list of expected Segments that make up this Resource ID
+	Segments() []Segment
+}
+
+type Segment struct {
+	// ExampleValue is an example of a value for this field, which is intended only to
+	// be used as a placeholder.
+	ExampleValue string
+
+	// FixedValue is the Fixed/Static value for this segment - only present when `Type`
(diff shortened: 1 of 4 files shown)


koikonom (approved):

mbfrahry (approved):
LGTM!

katbyte (approved):
Thanks @tombuildsstuff - LGTM 🏗️

PR #85, opened 2021-11-04
Introducing Common Resource ID's and Schema

tombuildsstuff:
This PR makes a number of changes to this codebase, namely this:

* Introduces Common Resource ID Types (using the new Resource ID Parser) for:
  * Management Group ID
  * Resource Group ID
  * Subscription ID
  * User Assigned Identity ID
* Introduces a new `lang` package for Golang specific helpers, this includes a `pointer` package (which is intentionally not called `to` since this conflicts with the pre-existing Azure/go-autorest package of the same name.
* Introduces Common Schema for:
  * Location
  * Resource Group Name
  * Identity (`SystemAssigned`, `SystemAssigned, UserAssigned` and `UserAssigned`)
  * Tags - note these Expand functions are only for the new tags, and intentionally not for the existing wrappers

There's also a couple of breaking changes to call out:

* The `formatting` package has been moved to `lang/formatting`
* The `response` package has been moved to `lang/response`

Whilst I've ported over most of the existing code, I've **intentionally not** ported over the following:

* `SchemaResourceGroupNameDiffSuppress` - ID Parsers should mean this is no longer necessary, so we can remove these piecemeal when that's done
* `SchemaResourceGroupNameDeprecatedComputed` - for behavioural consistency we should set these fields to an empty string so users know to remove these values


manicminer (approved):
LGTM! 🚀

lang/pointer/to.go
```diff
@@ -0,0 +1,33 @@
+package pointer
+
+func ToBool(input bool) *bool {
```
katbyte, line 3: and same here - this reads as "pointer . to bool ()" vs it being "pointer . from bool ()"

PR #86, opened 2021-11-10
resourcemanager/resourceids: raising an error if any segments are empty

tombuildsstuff:
This PR introduces a check at the end of the Resource ID parsing to ensure that all of the required segments have a value, which fixes an issue where the last segment could be empty and pass validation


jackofallops (approved):
LGTM 👍

PR #87, opened 2021-11-10
resourcemanager/identity: adding custom types for System Assigned / System Assigned User Assigned / User Assigned Identities

tombuildsstuff:
These are intentionally opinionated to match the behaviours both Azure expects (insofar as certain fields can't be marshalled) and that Terraform is expecting (insofar as when the `identity` block is omitted it should be set to `None` etc)

This means these intentionally use custom marshalling functions which is unfortunate but guarantees the behaviour here

This also removes _some of_ the superfluous `Schema`'s present, once this is in I'll go through and fix any remnants/the pointer package since this is already big enough


katbyte (approved):
LGTM 🧱

PR #90, opened 2021-11-29
Support for BearerAuthorizerCallback with MSAL tokens

manicminer:
Also, use object ID from token claims where possible instead of calling out to AAD Graph or MS Graph. Tokens should always have an `oid` claim but fallback is provided to query a graph API in the event unmarshaling fails or the claim is missing/empty.


authentication/config.go
```diff
@@ -100,9 +101,9 @@ func (c Config) BuildOAuthConfig(activeDirectoryEndpoint string) (*OAuthConfig,
 	return &multiAuth, nil
 }
 
-// BearerAuthorizerCallback returns a BearerAuthorizer valid only for the Primary Tenant
+// ADALBearerAuthorizerCallback returns a BearerAuthorizer valid only for the Primary Tenant
```
manicminer, line 104: Open to suggestions on naming these and/or whether we can break this exported method?

katbyte (approved):
LGTM 💯

PR #92, opened 2021-12-12
identity: introducing System OR User Assigned Identity Types

tombuildsstuff:
This PR introduces a new Identity type for System OR User Assigned Identities - this is because some API's implement System AND User Assigned Identities whilst others implement System OR User Assigned Identities.

Whilst we currently have support for System AND User Assigned - this PR introduces System OR User Assigned and so renames the existing types to System AND User Assigned, which whilst a breaking change is used in a handful of places so this seems fine.


resourcemanager/identity/system_or_user_assigned_list.go
```diff
@@ -0,0 +1,99 @@
...
+
+	out := map[string]interface{}{
+		"type":                   string(identityType),
+		"userAssignedIdentities": userAssignedIdentityIds,
+	}
+	return json.Marshal(out)
+}
+
+// ExpandSystemOrAssignedUserAssignedList expands the schema input into a SystemOrUserAssignedList struct
+func ExpandSystemOrAssignedUserAssignedList(input []interface{}) (*SystemOrUserAssignedList, error) {
```
jackofallops, line 46: Should this be?
```suggestion
func ExpandSystemOrUserAssignedList(input []interface{}) (*SystemOrUserAssignedList, error) {
```

jackofallops (approved):
One small note, but otherwise LGTM 👍

resourcemanager/identity/system_or_user_assigned_list.go
tombuildsstuff: 🤦 good spot, ta

PR #95, opened 2022-01-18
resourcemanager/tags: Flatten now returns map[string]interface{}

tombuildsstuff:
This enables this to be used directly in Read functions rather than having an intermediate translation function.


jackofallops (approved):
LGTM 👍

PR #96, opened 2022-01-18
New Common ID: Scope

tombuildsstuff:
Scope is used in a bunch of places, so this makes sense to have a Common Resource ID that we can reuse as necessary

This PR also updates the other Common Resource ID Parsers to the latest version


jackofallops (approved):
Thanks @tombuildsstuff - LGTM 👍

PR #97, opened 2022-01-19
commonschema: adding support for Extended Location

tombuildsstuff:
This is a kind of location which has more dynamic names than for the location field as such we'll have to leave this with lax validation for now.

### DIFF
--- a/resourcemanager/commonschema/location.go
+++ b/resourcemanager/commonschema/location.go
@@ -3,7 +3,19 @@ package commonschema
 import (
 	"github.com/hashicorp/go-azure-helpers/resourcemanager/location"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
+
+func ExtendedLocation() *schema.Schema {
+	return &schema.Schema{
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		StateFunc:        location.StateFunc,
+		DiffSuppressFunc: location.DiffSuppressFunc,
+		ValidateFunc:     validation.StringIsNotEmpty,
+	}
+}
 
 func Location() *schema.Schema {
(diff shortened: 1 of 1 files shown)


katbyte (approved):
LGTM 🏗️

resourcemanager/commonschema/location.go
```diff
@@ -3,8 +3,20 @@ package commonschema
 import (
 	"github.com/hashicorp/go-azure-helpers/resourcemanager/location"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
+func ExtendedLocation() *schema.Schema {
+	return &schema.Schema{
+		Type:             schema.TypeString,
+		Required:         true,
```
neil-yechenwei, line 12: Thanks @tombuildsstuff . Per my understanding, it's an optional property. If I am wrong, please correct me. Because it allows customers to take advantage of more granular locations when this property is set. If user doesn't set it, the resource would be deployed to normal location like "west europe". Does it make sense?

resourcemanager/commonschema/location.go
```diff
@@ -3,8 +3,20 @@ package commonschema
...
+		ForceNew:         true,
+		StateFunc:        location.StateFunc,
+		DiffSuppressFunc: location.DiffSuppressFunc,
+		ValidateFunc:     validation.StringIsNotEmpty,
```
neil-yechenwei, line 16: As there is a [list API](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list-locations) to retrieve available extended locations, but it has to be whitelisted by service team first before calling this API. So maybe here can add a property validation for extendedLocation like the [location validation](https://github.com/hashicorp/go-azure-helpers/blob/41ef58a3f0054aba44bb4231c2063682395fca58/resourcemanager/commonschema/location.go#L13) when this property is treated as optional property otherwise it would break existing users/resources. Does it make sense?

PR #98, opened 2022-01-25
lang/response: adding WasBadRequest

tombuildsstuff:
This has come up a number of times, so would be useful to have here


jackofallops (approved):
LGTM 👍

PR #99, opened 2022-02-01
identity: refactoring to be Required, Optional and Computed

tombuildsstuff:
We need to support where Identity is both Required, Optional and Computed - however whilst doing so I realised that adding a `{Name}Required` function would differ from the schema used in Resource Groups and Location where the Required method is the default, so this PR refactors the identity function to take this into account.

### DIFF
--- a/resourcemanager/commonschema/identity_system.go
+++ b/resourcemanager/commonschema/identity_system.go
@@ -7,7 +7,36 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
-func SystemAssignedIdentity() *schema.Schema {
+// SystemAssignedIdentityRequired returns the System Assigned Identity schema where this is Required
+func SystemAssignedIdentityRequired() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Required: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"type": {
+					Type:     schema.TypeString,
+					Required: true,
+					ValidateFunc: validation.StringInSlice([]string{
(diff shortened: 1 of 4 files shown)


resourcemanager/commonschema/identity_system.go
```diff
@@ -7,7 +7,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
-func SystemAssignedIdentity() *schema.Schema {
+// SystemAssignedIdentityRequired returns the System Assigned Identity schema where this is Required
+func SystemAssignedIdentityRequired() *schema.Schema {
 	return &schema.Schema{
 		Type:     schema.TypeList,
 		Optional: true,
```
jackofallops, line 14: Should this be `Required` ?

resourcemanager/commonschema/identity_system_user.go
```diff
@@ -13,7 +13,8 @@ import (
 // from a users perspective however, these should both be represented using the same schema
 // so we have a single schema and separate Expand/Flatten functions
 
-func SystemAssignedUserAssignedIdentity() *schema.Schema {
+// SystemAssignedUserAssignedIdentityRequired returns the System Assigned User Assigned Identity schema where this is Required
+func SystemAssignedUserAssignedIdentityRequired() *schema.Schema {
 	return &schema.Schema{
 		Type:     schema.TypeList,
 		Optional: true,
```
jackofallops, line 20: Same here, `Required`?

jackofallops (approved):
2 questions below, otherwise LGTM 👍

resourcemanager/commonschema/identity_system.go
tombuildsstuff: Yes, I thought I reset something - ta.

PR #100, opened 2022-02-04
commonschema/identity: adding ForceNew options as some are Required/Optional and ForceNew

tombuildsstuff:


jackofallops (approved):
Thanks @tombuildsstuff  - LGTM 👍

PR #101, opened 2022-02-07
resourcemanager/identity: normalizing the casing of the `type`

tombuildsstuff:
This works around an issue where the API specifies one case but returns another, such as in https://github.com/Azure/azure-rest-api-specs/issues/12330 - as such these should be parsed insensitively when coming back from the API


manicminer (approved):
LGTM 🪑

jackofallops (approved):
LGTM 👍

PR #102, opened 2022-02-09
Identity: add struct tags for HCL decoding

manicminer:


resourcemanager/identity/system_and_user_assigned_list.go
```diff
@@ -3,17 +3,18 @@ package identity
...
 
 type SystemAndUserAssignedList struct {
-	Type        Type     `json:"type"`
-	PrincipalId string   `json:"principalId"`
-	TenantId    string   `json:"tenantId"`
-	IdentityIds []string `json:"userAssignedIdentities"`
+	Type        Type     `json:"type" tfschema:"type"`
+	PrincipalId string   `json:"principalId" tfschema:"principal_id"`
+	TenantId    string   `json:"tenantId" tfschema:"tenant_id"`
+	IdentityIds []string `json:"userAssignedIdentities" tfschema:"identity_ids"`
```
tombuildsstuff, line 17: does this work? These are intentionally a Set rather than a List in the common schema, and iirc right now we only deserialize a List?

resourcemanager/identity/system_and_user_assigned_map.go
```diff
@@ -11,10 +11,10 @@ import (
...
 
 type SystemAndUserAssignedMap struct {
-	Type        Type                                   `json:"type"`
-	PrincipalId string                                 `json:"principalId"`
-	TenantId    string                                 `json:"tenantId"`
-	IdentityIds map[string]UserAssignedIdentityDetails `json:"userAssignedIdentities"`
+	Type        Type                                   `json:"type" tfschema:"type"`
+	PrincipalId string                                 `json:"principalId" tfschema:"principal_id"`
+	TenantId    string                                 `json:"tenantId" tfschema:"tenant_id"`
+	IdentityIds map[string]UserAssignedIdentityDetails `json:"userAssignedIdentities" tfschema:"identity_ids"`
```
tombuildsstuff, line 17: these are exposed in the schema as a list of ID's currently (the other metadata we may want to expose in the future, which is why it's plumbed through, fwiw) - so I don't think we can use this value directly right now?

resourcemanager/identity/system_or_user_assigned_list.go
```diff
@@ -11,10 +11,10 @@ import (
...
 
 type SystemOrUserAssignedList struct {
-	Type        Type     `json:"type"`
-	PrincipalId string   `json:"principalId"`
-	TenantId    string   `json:"tenantId"`
-	IdentityIds []string `json:"userAssignedIdentities"`
+	Type        Type     `json:"type" tfschema:"type"`
+	PrincipalId string   `json:"principalId" tfschema:"principal_id"`
+	TenantId    string   `json:"tenantId" tfschema:"tenant_id"`
+	IdentityIds []string `json:"userAssignedIdentities" tfschema:"identity_ids"`
```
tombuildsstuff, line 17: (as above)

resourcemanager/identity/system_or_user_assigned_map.go
```diff
@@ -11,10 +11,10 @@ import (
...
 
 type SystemOrUserAssignedMap struct {
-	Type        Type                                   `json:"type"`
-	PrincipalId string                                 `json:"principalId"`
-	TenantId    string                                 `json:"tenantId"`
-	IdentityIds map[string]UserAssignedIdentityDetails `json:"userAssignedIdentities"`
+	Type        Type                                   `json:"type" tfschema:"type"`
+	PrincipalId string                                 `json:"principalId" tfschema:"principal_id"`
+	TenantId    string                                 `json:"tenantId" tfschema:"tenant_id"`
+	IdentityIds map[string]UserAssignedIdentityDetails `json:"userAssignedIdentities" tfschema:"identity_ids"`
```
tombuildsstuff, line 17: (as above)

resourcemanager/identity/user_assigned_list.go
```diff
@@ -4,16 +4,15 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 var _ json.Marshaler = &UserAssignedList{}
 
 type UserAssignedList struct {
-	Type        Type     `json:"type"`
-	IdentityIds []string `json:"userAssignedIdentities"`
+	Type        Type     `json:"type" tfschema:"type"`
+	IdentityIds []string `json:"userAssignedIdentities" tfschema:"identity_ids"`
```
tombuildsstuff, line 15: (as above)

resourcemanager/identity/user_assigned_map.go
```diff
@@ -11,8 +11,8 @@ import (
 var _ json.Marshaler = &UserAssignedMap{}
 
 type UserAssignedMap struct {
-	Type        Type                                   `json:"type"`
-	IdentityIds map[string]UserAssignedIdentityDetails `json:"userAssignedIdentities"`
+	Type        Type                                   `json:"type" tfschema:"type"`
+	IdentityIds map[string]UserAssignedIdentityDetails `json:"userAssignedIdentities" tfschema:"identity_ids"`
```
tombuildsstuff, line 15: (as above)

tombuildsstuff (changes requested):
hey @manicminer 

Thanks for this PR - taking a look through whilst this is looking good I think there's an issue here when serializing/deserializing the `identity_ids` properties, since these are (intentionally) a Set of strings which we currently don't deserialize iirc?

resourcemanager/identity/system_and_user_assigned_list.go
manicminer: I believe decoding [does work](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/sdk/resource_decode.go#L127-L129)? Although it doesn't have test coverage. I [gave it a try](https://github.com/hashicorp/terraform-provider-azurerm/commit/0d1c8f6532f097a6912f7eedce813435690c91c6) and it seems ok.

Encoding should also work, since it's only concerned with the values passed to d.Set() - in this case would be a slice regardless of whether the underlying type is a List or a Set?

resourcemanager/identity/system_and_user_assigned_map.go
manicminer: Ah ok, I guess we can take this one out for now until it's ready

katbyte (approved):
As you've checked LGTM 🧊

PR #106, opened 2022-03-03
resourcemanager/identity: introducing a struct/expand/flatten functions for the V3 (Legacy) Identity Types

tombuildsstuff:
This ensures the UX is consistent insofar as the legacy value `SystemAssigned,UserAssigned` becomes consistent
as `SystemAssigned, UserAssigned`


jackofallops (approved):
LGTM 👍

PR #108, opened 2022-03-16
resourcemanager/commonschema: adding support for Edge Zones

tombuildsstuff:
These are intended to be used as `edge_zone = commonschema.EdgeZoneOptional()`

At this time we intentionally only support Optional, OptionalForceNew and Computed as these are the use-cases - however we may expand this to support Required in future.

Supersedes #97


manicminer (approved):
LGTM 🚢

jackofallops (approved):
LGTM 🚀

PR #109, opened 2022-03-25
Bugfix: add compatiblity for ADAL tokens in auth.Config.MSALBearerAuthorizerCallback()

manicminer:
Ensure that tokens returned via Azure CLI or MSI authentication will still work for data plane APIs, because those both return legacy ADAL tokens.


katbyte (approved):
LGTM 🌻

PR #112, opened 2022-04-24
Identity: Add expand/flatten from/to the typed schema model

magodo:
The identity package currently has support for expanding/flattening from/to the raw representation of identity (i.e. `[]interface{}`), to/from the SDK types (e.g. `SystemAndUserAssignedList`). This works fine for the untyped flavor code. However, it lacks of supports for the code using typed SDK.

This PR defines models to be used for the typed SDK:

- `ModelUserAssinged`: for user assigned identity (either the user assigned identity in the SDK definition is a list or a map)
- `ModeSystemAssigned`: for system assigned identity
- `ModeSystemAssignedUserAssigned`: for system and/or user assigned identity, it doesn't matter whether it is "or" or "and" for the schema model (thought it matters when choosing the correct schema and the correct expand/flatten function).

Arguably, the `tfschema` tags can be removed for the SDK types, as they are not expected to be used as part of the schema model, instead they represents the SDK type. But if we directly remove them, it will affect the current provider code for following files:

- internal/services/mssql/mssql_managed_instance_resource.go
- internal/services/mssql/mssql_managed_instance_data_source.go

Once this PR got merged, we can do a follow up PR to change these files to using the typed model, and remove the `tfschema` tags for the SDK types.

### DIFF
--- a/resourcemanager/identity/system_or_user_assigned_map.go
+++ b/resourcemanager/identity/system_or_user_assigned_map.go
@@ -132,8 +132,8 @@ func ExpandSystemOrUserAssignedMapFromModel(input []ModelSystemAssignedUserAssig
 	}, nil
 }
 
-// FlattenSystemOrUserAssignedMapFromModel turns a SystemOrUserAssignedMap into a typed schema model
-func FlattenSystemOrUserAssignedMapFromModel(input *SystemOrUserAssignedMap) (*[]ModelSystemAssignedUserAssigned, error) {
+// FlattenSystemOrUserAssignedMapToModel turns a SystemOrUserAssignedMap into a typed schema model
+func FlattenSystemOrUserAssignedMapToModel(input *SystemOrUserAssignedMap) (*[]ModelSystemAssignedUserAssigned, error) {
 	if input == nil {
 		return &[]ModelSystemAssignedUserAssigned{}, nil
 	}
--- a/resourcemanager/identity/system_or_user_assigned_map.go
+++ b/resourcemanager/identity/system_or_user_assigned_map.go
@@ -110,7 +110,7 @@ func ExpandSystemOrUserAssignedMapFromModel(input []ModelSystemAssignedUserAssig
 	if len(input) == 0 {
 		return &SystemOrUserAssignedMap{
 			Type:        TypeNone,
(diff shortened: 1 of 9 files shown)


resourcemanager/identity/user_assigned_map.go
```diff
@@ -92,3 +92,56 @@ func FlattenUserAssignedMap(input *UserAssignedMap) (*[]interface{}, error) {
 		},
 	}, nil
 }
+
+// ExpandUserAssignedMapFromModel expands the typed schema input into a UserAssignedMap struct
+func ExpandUserAssignedMapFromModel(input []ModelUserAssigned) (*UserAssignedMap, error) {
+	if len(input) == 0 {
+		return &UserAssignedMap{
+			Type:        TypeNone,
+			IdentityIds: map[string]UserAssignedIdentityDetails{},
```
tombuildsstuff, line 101: pedantic, but we should probably remove this to be `nil` rather than an empty map (even if they're semantically the same):

```suggestion
```

resourcemanager/identity/system_or_user_assigned_map.go
```diff
@@ -104,3 +104,60 @@ func FlattenSystemOrUserAssignedMap(input *SystemOrUserAssignedMap) (*[]interfac
 		},
 	}, nil
 }
+
+// ExpandSystemOrUserAssignedMapFromModel expands the typed schema input into a SystemOrUserAssignedMap struct
+func ExpandSystemOrUserAssignedMapFromModel(input []ModelSystemAssignedUserAssigned) (*SystemOrUserAssignedMap, error) {
+	if len(input) == 0 {
+		return &SystemOrUserAssignedMap{
+			Type:        TypeNone,
+			IdentityIds: map[string]UserAssignedIdentityDetails{},
```
tombuildsstuff, line 113: (same here)

tombuildsstuff, line 113: (same here)

tombuildsstuff (approved):
Left a few comments inline but this otherwise LGTM 👍

magodo:
Hey @tombuildsstuff, sorry for missing this one. I've resolved the comments, please take another look!

tombuildsstuff (approved):

PR #113, opened 2022-04-25
Upgrading to Go 1.18 / adding an interface for Normalizers

tombuildsstuff:
A Normalizer transforms a value if possible from whatever input form to the expected/normalized form - this is helpful when the casing returned from the API may differ from the casing we're expecting.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/hashicorp/go-azure-helpers
 
+go 1.18
+
 require (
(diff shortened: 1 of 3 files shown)


jackofallops (changes requested):
Thanks @tombuildsstuff - One change I believe needed below, will catch up with you offline to discuss.

tombuildsstuff:
This PR's no longer needed - closing.

Issue #91, opened 2021-12-02
New Authentication Method: OIDC


tjololo:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently the only way to authenticate with a service principal you need to use secert or certificate. It would be nice if the azurerm provider supports authenticating a Service Principal or a AzureAD App using the Azure CLI.

Use-cases
Github announced OIDC Federated credentials for Github Actions. It can be used by the Azure Login action. It would be nice if azurerm also supported that scenario so you do not need a secret for deployment to Azure anymore.

See https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure for details.

Issue was raised with hashicorp/terraform earlier: [#29830](https://github.com/hashicorp/terraform/issues/29830)

### New or Affected Resource(s)

No resources should be affected, this only a new way to authenticate with a service principal

### Potential Terraform Configuration

No changes to the terraform configuration as authentication is handled by Azure CLI

### References

* [hashicorp/terraform/issue#29830](https://github.com/hashicorp/terraform/issues/29830)

dhduvall:
This seems related to hashicorp/terraform-provider-azurerm#5271, which talks explicitly about JWT authentication, but I think this is asking for the same thing.

GitLab also issues a [signed JWT for its CI runs](https://docs.gitlab.com/ee/ci/examples/authenticating-with-hashicorp-vault/), and I'd like to be able to use that to authenticate to Azure. I guess the workaround is to have a pre-build step that exchanges the JWT for a secret, and then makes that secret available to Terraform, but I haven't tried it yet.

Note that in this case, the need for a refresh mechanism as mentioned in the other issue isn't required, since the JWT expiration is matched to the timeout of the job.

tombuildsstuff:
@dhduvall whilst we're looking into OIDC we have no plans to support JWT authentication - so that's out-of-scope here unfortunately.

SudoSpartanDan:
[...]

PR #114, opened 2022-04-25
Support OIDC Authentication

manicminer:
Closes: #91


manicminer:
Test results

Screenshot 2022-04-25 at 21 35 04

authentication/auth_method_oidc.go
```diff
@@ -0,0 +1,93 @@
...
+		ClientID:            a.clientId,
+		IDTokenRequestURL:   a.idTokenRequestUrl,
+		IDTokenRequestToken: a.idTokenRequestToken,
+		Scopes:              []string{api.DefaultScope()},
+	}
+
+	return &authWrapper.Authorizer{Authorizer: conf.TokenSource(ctx)}, nil
+}
+
+func (a oidcAuth) populateConfig(c *Config) error {
```
tombuildsstuff, line 65: this'll need to be a pointer:

```suggestion
func (a *oidcAuth) populateConfig(c *Config) error {
```

authentication/auth_method_oidc_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func TestOIDC_populateConfig(t *testing.T) {
+	config := &Config{}
+	err := oidcAuth{}.populateConfig(config)
+	if err != nil {
+		t.Fatalf("Error populating config: %s", err)
+	}
+	if !config.AuthenticatedAsAServicePrincipal {
+		t.Fatalf("Expected `AuthenticatedAsAServicePrincipal` to be true but it wasn't")
+	}
```
tombuildsstuff, line 89: given we aren't guaranteed a refresh token when using OIDC, I think we could do with a flag to know this is OIDC to handle that under certain conditions?

tombuildsstuff (approved):
Left a few comments inline but otherwise 👍 nice one

authentication/auth_method_oidc_test.go
manicminer: We automatically get a new ID token from GitHub and use that to get a new access token to essentially fake a refresh - if that fails then there are essentially no recovery options to use anyway.

PR #115, opened 2022-04-26
Force use of Microsoft Graph when using OIDC authentication

manicminer:


authentication/auth_method_oidc.go
```diff
@@ -65,6 +65,7 @@ func (a oidcAuth) getMSALToken(ctx context.Context, api environments.Api, _ auto
 func (a oidcAuth) populateConfig(c *Config) error {
 	c.AuthenticatedAsAServicePrincipal = true
 	c.AuthenticatedViaOIDC = true
+	c.UseMicrosoftGraph = true
```
tombuildsstuff, line 68: I think we can remove this in favour of adding it to the `isApplicable` function above?

tombuildsstuff (changes requested):
I think this wants moving to the `isApplicable` function?

katbyte (approved):
LGTM 🌻

PR #117, opened 2022-05-24
Add SystemData struct.

koikonom:
This is a simple read only structure. It does not specify a custom unmarshal function, and the marshal function will always return an empty string.

This PR adds functionality requited by https://github.com/hashicorp/pandora/pull/808

### DIFF
--- a/resourcemanager/systemdata/model.go
+++ b/resourcemanager/systemdata/model.go
@@ -1,0 +1,18 @@
+package systemdata
+
+import "encoding/json"
+
+var _ json.Unmarshaler = &SystemData{}
+
+type SystemData struct {
+	CreatedBy          string `json:"createdBy"`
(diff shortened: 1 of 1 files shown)


tombuildsstuff (approved):
LGTM 👍

resourcemanager/systemdata/model.go
```diff
@@ -0,0 +1,20 @@
+package systemdata
+
+import "encoding/json"
+
+var _ json.Marshaler = &SystemAssigned{}
```
tombuildsstuff, line 5:
```suggestion
var _ json.Marshaler = &SystemData{}
```

tombuildsstuff (changes requested):
Spotted whilst hitting approve 🙃

PR #119, opened 2022-05-30
Auth: CLI method mitigate breaking change introduced in CLI v2.37.0

### LINKED ISSUE #118: Auth: `populateConfig` for CLI is missing `object_id` since CLI v2.37.0
Since v2.37.0, CLI migrated the underlying API for `az ad` to using MS graph: https://github.com/Azure/azure-cli/pull/22432. This causes different format for the output of `az ad signed-in-user show -o=json`.

Related issue: https://github.com/hashicorp/terraform-provider-azurerm/issues/16982

magodo:
Since v2.37.0, CLI migrates to using MS graph for the AD related apis,
this causes breaking change on the resulting data model, which includes
feild `object_id` being renamed to `id`. This causes CLI auth method
failed to work for CLI since v2.37.0.

This commit checks the CLI version and read the corresponding field for
object id differently, for both CLI and multi tenant CLI auth methods.

Besides, the multi tenant CLI auth method seems to be broken even prior
to v2.37.0, when there is nothing configured in the provider block. This
commit also fixed that part by populating the tenant id and subscription
id in case they are empty.

Fix #118

### DIFF
--- a/authentication/auth_method_azure_cli_token.go
+++ b/authentication/auth_method_azure_cli_token.go
@@ -25,6 +25,8 @@ type azureCLIProfile struct {
 	subscriptionId string
 	tenantId       string
 	tenantOnly     bool
+
+	azVersion version.Version
 }
 
 type azureCliTokenAuth struct {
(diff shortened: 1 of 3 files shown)


authentication/auth_method_azure_cli_token.go
```diff
@@ -25,6 +25,8 @@ type azureCLIProfile struct {
 	subscriptionId string
 	tenantId       string
 	tenantOnly     bool
+
+	azVersion version.Version
```
tombuildsstuff, line 29: we can't do it today since it's only just rolled out, but in time we should bump the minimum version to v2.37 since ADAL support will be entirely gone - WDYT @manicminer?

tombuildsstuff (approved):
One thought (for the future) whilst passing through, but this otherwise LGTM 👍

authentication/auth_method_azure_cli_token.go
manicminer: Yeah that makes sense. Probably best to maintain support for older versions until the API is sunset, or the next provider major (whichever comes first)?

manicminer (approved):
Thanks for this fix @magodo, this LGTM 👍

authentication/auth_method_azure_cli_token.go
tombuildsstuff: once there's an end-date published for the older method I guess we can update it then?

PR #120, opened 2022-06-09
Generic implementation for the `pointer`/`tfdata` `ToXXX()` and `FromXXX()` utilities

magodo:
For `pointer` package, rewrite the existing functions via generics.

Adding a new package `tfdata` to rewrite the same thing from https://github.com/hashicorp/terraform-provider-azurerm/blob/main/utils/common_marshal.go, using generics (except one for the delimiter string, which is not needed to be generics).

### DIFF
--- a/lang/pointer/to.go
+++ b/lang/pointer/to.go
@@ -74,3 +74,32 @@ func ToString(input *string) string {
 
 	return ""
 }
+
+type primary interface {
+	~bool | ~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~float32 | ~float64 | ~string
+}
+
+// ToPrimary turns a pointer to a primary into a primary, returning its zero value if it's nul
+func ToPrimary[T primary](input *T) T {
+	if input != nil {
+		return *input
+	}
+	var v T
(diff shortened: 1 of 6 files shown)


jackofallops:
Hi @magodo - Thanks for this PR. At this time whilst we appreciate the effort, we’re still trying to work out exactly how Generics could be useful in the context of the Provider. Given the number of structural changes currently in flight, we’ll need to put this on hold for the moment until the associated tooling (such as linters) is updated to support generics, and we've had chance to investigate how we want to take this forward.

Thanks again.

lang/pointer/to.go
```diff
@@ -74,3 +76,28 @@ func ToString(input *string) string {
...
+// ToMap turns a pointer to a map into a map, returning its zero value if it's nul
+func ToMap[K types.Primary, T types.Primary | interface{}](input *map[K]T) map[K]T {
+	if input != nil {
+		return *input
+	}
+	return map[K]T{}
+}
+
+// ToSlice turns a pointer to a slice into a slice, returning its zero value if it's nul
+func FromSlice[T types.Primary | interface{}](input *[]T) []T {
```
magodo, line 98: A typo here.

jackofallops:
Closing as already implemented.

PR #122, opened 2022-06-29
Poller: add a new poller with method in constructor and check for 404 responses

stephybun:
Related to https://github.com/hashicorp/pandora/issues/961

### DIFF
--- a/polling/poller.go
+++ b/polling/poller.go
@@ -3,7 +3,9 @@ package polling
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/go-azure-helpers/lang/response"
 	"net/http"
+	"strings"
 
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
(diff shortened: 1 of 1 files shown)


polling/poller.go
```diff
@@ -3,7 +3,9 @@ package polling
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/go-azure-helpers/lang/response"
```
tombuildsstuff, line 6: import-ordering

polling/poller.go
```diff
@@ -32,13 +37,33 @@ func NewLongRunningPollerFromResponse(ctx context.Context, resp *http.Response,
...
 	if fw.future == nil {
 		return fmt.Errorf("internal error: cannot poll on a nil-future")
 	}
 
 	err := fw.future.WaitForCompletionRef(fw.ctx, fw.client)
+	if strings.EqualFold(fw.method, "DELETE") {
+		if !response.WasNotFound(fw.future.Response()) {
+			return err
+		}
+	}
```
tombuildsstuff, line 66: can we move this down a line? whilst a different choice to the provider generally, having the `HttpResponse` available can be useful should we need to override things

polling/poller.go
```diff
@@ -32,13 +37,33 @@ func NewLongRunningPollerFromResponse(ctx context.Context, resp *http.Response,
...
 // PollUntilDone polls until this Long Running Poller is completed
 func (fw *LongRunningPoller) PollUntilDone() error {
 	if fw.future == nil {
 		return fmt.Errorf("internal error: cannot poll on a nil-future")
 	}
 
 	err := fw.future.WaitForCompletionRef(fw.ctx, fw.client)
+	if strings.EqualFold(fw.method, "DELETE") {
+		if !response.WasNotFound(fw.future.Response()) {
+			return err
```
tombuildsstuff, line 64: since we're returning `err` below anyway, we could invert this to become:

```suggestion
		if response.WasNotFound(fw.future.Response()) {
			err = nil
```

tombuildsstuff (approved):
LGTM 👍

PR #123, opened 2022-07-11
Storage Account SAS: The format is changed if signed version >= 2020-12-06

magodo:
This is related to https://github.com/hashicorp/terraform-provider-azurerm/issues/17558, https://github.com/hashicorp/terraform-provider-azurerm/issues/18807

See https://docs.microsoft.com/en-us/rest/api/storageservices/create-account-sas#construct-the-signature-string for reference.

### DIFF
--- a/storage/sas_token.go
+++ b/storage/sas_token.go
@@ -19,6 +19,7 @@ const (
 
 // ComputeAccountSASToken computes the SAS Token for a Storage Account based on the
 // access key & given permissions
+// See: https://docs.microsoft.com/en-us/rest/api/storageservices/create-account-sas
 func ComputeAccountSASToken(accountName string,
 	accountKey string,
 	permissions string,
@@ -29,6 +30,7 @@ func ComputeAccountSASToken(accountName string,
 	signedProtocol string,
 	signedIp string, // nolint: unparam
 	signedVersion string, // nolint: unparam
+	signedEncryptionScope string, // nolint: unparam
 ) (string, error) {
 
 	// UTF-8 by default...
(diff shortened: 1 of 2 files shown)


storage/sas_token.go
```diff
@@ -42,6 +44,10 @@ func ComputeAccountSASToken(accountName string,
 	stringToSign += signedProtocol + "\n"
 	stringToSign += signedVersion + "\n"
 
+	if signedVersion >= "2020-12-06" {
```
tombuildsstuff, line 47: since this is a string this comparison isn't guaranteed to work? might be worth conditionally appending this field if the users specified it, and then calling this out in the documentation instead?

magodo: If the version string is guaranteed to be consisten on the format, then the comparison should work as expected. The field can't be appended conditionally, e.g. for version equal or later than 2020-12-06, even for empty `signedEncryptionScope`, the new line is necessary to be appended, otherwise the SAS will be rejected by the service.

westonplatter:
@magodo much appreciation for opening the PR. I have a need for this fix on some of my infra. Is there anything I can do to help move along the PR?

jefflaub12:
@tombuildsstuff Any update on delivery for this fix?

storage/sas_token.go
westonplatter: @tombuildsstuff tested this in go playground, and @magodo is correct (go version = 1.19 and 1.20)


image

bartvanackooij:
Hi, 

Any update on this? Anything we can do to help this get this fixed?

isaacfraser-tomtom:
Also encountering this issues and hoping for this to be fixed

westonplatter:
@tombuildsstuff I realize you conducted the first review of these changes. Are you able to continue reviewing the pull request? Please let me know if there's any way I can assist or support in getting these changes merged. I have production infrastructure that require these changes at my full-time job and therefore have ample time to give to developing a solution to the issue.

storage/sas_token.go
tombuildsstuff: @magodo apologies for the super delayed re-review here - I'd missed this one hadn't been re-reviewed/merged - if we can add a unit test to check this one it should otherwise be good to merge

magodo:
@tombuildsstuff Thanks for the review! I've added up a unit test to cover this new `sv`. Meanwhile, I've verified that when using the same `sv` and empty `ses`, the code here generates the same SAS as is generated from Portal.

PR #124, opened 2022-07-11
lang/response: making WasNotFound public

tombuildsstuff:
This allows custom status codes to be checked as required to workaround Swagger/API issues whilst also handling the response being nil due to a dropped connection or similar.


jackofallops (approved):
LGTM 👍

PR #125, opened 2022-07-13
Support a default environment from metadata

manicminer:
Support a default environment when a single env is returned from the metadata service, and no environment name was specified


katbyte (approved):
LGTM 🎩

PR #127, opened 2022-08-11
add common ids supporting the Network RP

jackofallops:


tombuildsstuff (approved):
Two comments about the file name but this otherwise LGTM 👍

PR #129, opened 2022-08-16
CLI auth: refresh account list prior to getting subscription id during build

magodo:
Refresh account list prior to getting subscription id during building the auth.

Related to https://github.com/hashicorp/terraform-provider-azurerm/issues/17979

### DIFF
--- a/authentication/auth_method_azure_cli_token.go
+++ b/authentication/auth_method_azure_cli_token.go
@@ -50,6 +50,10 @@ func (a azureCliTokenAuth) build(b Builder) (authMethod, error) {
 		servicePrincipalAuthDocsLink: b.ClientSecretDocsLink,
 	}
 
+	if err := refreshAccounts(); err != nil {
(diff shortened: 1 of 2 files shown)


manicminer:
Thanks for this @magodo. Whilst I understand this is meant to help sync the subscription list when a user creates a new subscription dynamically - if hypothetically we moved in future to require a subscription ID be set in the AzureRM provider config, would this fix be negated?

magodo:
@manicminer I think our ultimate goal is to allow the provider's `subscription_id` be unknown until apply time (as the example given in https://github.com/hashicorp/terraform-provider-azurerm/issues/17979). If that is the goal, then we still need this change if both provider aliases are using Azure CLI auth. As after the first one created a new subscription, there will needs a "refresh" via Azure CLI before we can continue using the new subscription for the second provider alias.

manicminer:
@magodo Thanks for the clarification - I'll comment further on the azurerm issue to keep the conversation in one place, and we'll leave this open for now until we have agreed on a way forward. Thanks!

PR #131, opened 2022-09-09
Add ability to specify path to OIDC token

mpminardi:
Adds the ability to specify a path to an OIDC token in place of supplying the contents of the token.

### DIFF
--- a/authentication/auth_method_oidc.go
+++ b/authentication/auth_method_oidc.go
@@ -3,6 +3,7 @@ package authentication
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/hashicorp/go-multierror"
@@ -16,6 +17,7 @@ type oidcAuth struct {
 	clientId            string
 	environment         string
 	idToken             string
+	idTokenPath         string
 	idTokenRequestToken string
 	idTokenRequestUrl   string
 	tenantId            string
@@ -27,6 +29,7 @@ func (a oidcAuth) build(b Builder) (authMethod, error) {
 		clientId:            b.ClientID,
 		environment:         b.Environment,
 		idToken:             b.IDToken,
+		idTokenPath:         b.IDTokenPath,
 		idTokenRequestUrl:   b.IDTokenRequestURL,
 		idTokenRequestToken: b.IDTokenRequestToken,
 		tenantId:            b.TenantID,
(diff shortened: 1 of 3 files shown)


authentication/auth_method_oidc.go
```diff
@@ -65,19 +68,39 @@ func (a oidcAuth) getMSALToken(ctx context.Context, api environments.Api, _ auto
 		return &authWrapper.Authorizer{Authorizer: conf.TokenSource(ctx)}, nil
 	}
 
+	idToken := a.idToken
+
+	if a.idTokenPath != "" {
```
mpminardi, line 73: Not sure on priority of the token path vs actual token here, or if we would want to error out when both are supplied instead.

authentication/auth_method_oidc.go
```diff
@@ -27,6 +29,7 @@ func (a oidcAuth) build(b Builder) (authMethod, error) {
 		clientId:            b.ClientID,
 		environment:         b.Environment,
 		idToken:             b.IDToken,
+		idTokenPath:         b.IDTokenPath,
```
katbyte, line 32:
```suggestion
		idTokenFilePath:         b.IDTokenFilePath,
```

katbyte (changes requested):
Thanks @mpminardi - couple minor comments i've left inline but otherwise looks good

authentication/auth_method_oidc.go
```diff
@@ -65,19 +68,43 @@ func (a oidcAuth) getMSALToken(ctx context.Context, api environments.Api, _ auto
...
+
+	if a.idTokenFilePath != "" {
+		idToken, err = a.readTokenFile(a.idTokenFilePath)
+
+		if err != nil {
+			return nil, fmt.Errorf("reading token file: %v", err)
+		}
+
+		if a.idToken != "" && a.idToken != idToken {
+			return nil, fmt.Errorf("mismatch between supplied OIDC token and supplied OIDC token file")
```
mpminardi, line 81: @katbyte have added an error when the token doesn't match what is read from the token file based on your recommendation in https://github.com/hashicorp/go-azure-helpers/pull/131#discussion_r980315532. Let me know if you think this text should be adjusted at all!

katbyte: would it be a good idea to output the tokens?

katbyte, line 81: ie
```suggestion
			return nil, fmt.Errorf("mismatch between supplied OIDC token (%s) and supplied OIDC token file contents (%s) - please either remove one or ensure they match", a.idToken, idToken)
```

manicminer: I'm not sure that we should to be honest - we go to pains to exclude tokens from log output in other places.

manicminer (approved):
Thanks for the contribution and changes @mpminardi, this LGTM!

PR #133, opened 2022-10-04
refactor: resourceGroup -> resourceGroupName for consistency

tombuildsstuff:
These should be `resourceGroupName` rather than `resourceGroup`


mbfrahry (approved):
LGTM

PR #134, opened 2022-10-11
commonschema: adding types for Resource ID References

tombuildsstuff:
Today these are simply helper types, but over time these will allow the Resource IDs to be validated by differentiating as required on the segment types.


jackofallops (approved):
Thanks @tombuildsstuff - just one super minor thing below, otherwise LGTM 👍

PR #142, opened 2023-01-05
Container SAS token: don't urlencode the `se` and `st` (same as Portal)

magodo:
Relating to https://github.com/hashicorp/terraform-provider-azurerm/issues/19828


tombuildsstuff (approved):
LGTM - apologies for the delayed review on this one @magodo

PR #144, opened 2023-02-01
`auth` - support authentication with a bear token string

ms-zhenhua:
We have a requirement of authenticating with a bear token string when using AzureRM Provider.

### DIFF
--- a/authentication/auth_method_azure_bear_token.go
+++ b/authentication/auth_method_azure_bear_token.go
@@ -1,0 +1,64 @@
+package authentication
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hi @ms-zhenhua 

Thanks for this PR.

This has come up previously [here](https://github.com/hashicorp/terraform-provider-azurerm/issues/5271#issuecomment-611405401) and isn't something we can reliably use since there's no means of refreshing this token once it becomes expired - could you elaborate on the use-case here?

Thanks!

PR #147, opened 2023-02-02
resourcemanager/commonids: refactoring User Assigned Identity ID to use the updated segment name

tombuildsstuff:
Matching the behaviour introduced in https://github.com/hashicorp/pandora/pull/2011 which unblocks fixing https://github.com/hashicorp/pandora/actions/runs/4073772510


jackofallops (approved):
LGTM 👍

PR #149, opened 2023-02-05
Remove authentication features in favor of go-azure-sdk/sdk/auth

manicminer:
https://github.com/hashicorp/terraform-provider-azurerm/pull/20320 removes use of the go-azure-helpers/authentication package in favor of the new SDK's own auth package, except for the two remaining helper functions which are left in here.

~Additionally, rework the helper functions for obtaining the authenticated principal's object ID~

~Depends on https://github.com/manicminer/hamilton/pull/217 before merging~


go.mod
```diff
@@ -3,18 +3,14 @@ module github.com/hashicorp/go-azure-helpers
 require (
 	github.com/Azure/azure-sdk-for-go v66.0.0+incompatible
 	github.com/Azure/go-autorest/autorest v0.11.28
-	github.com/Azure/go-autorest/autorest/adal v0.9.22
-	github.com/Azure/go-autorest/autorest/azure/cli v0.4.6
-	github.com/hashicorp/go-multierror v1.1.1
-	github.com/hashicorp/go-version v1.6.0
+	github.com/hashicorp/go-azure-sdk v0.20230214.1122757-0.20230214215836-94fd7f7d8e2a
```
tombuildsstuff, line 6: I think vendoring this both directions may cause us issues.. is it worth inlining this logic within `azurerm` and `azuread` temporarily?

manicminer: That's a good call, I was also starting to get concerned about that. I'll code an inline client here using the new base layer.

tombuildsstuff (approved):

PR #153, opened 2023-02-25
`AzureEnvironmentByNameFromEndpoint`|`IsEnvironmentAzureStack` not restrict the `metadataHost` only to be a hostname

magodo:
Previously, the `endpoint` in the function `AzureEnvironmentByNameFromEndpoint` and `IsEnvironmentAzureStack` represents only the host name, and fix the request to the metadata host to using https. This makes customized metadata host (mostly running in localhost) has to run in HTTPS, which is not convenient sometimes. This PR relax that restriction to supports either a host name or a host name with scheme specified, in a non-breaking way.

### DIFF
--- a/authentication/environment.go
+++ b/authentication/environment.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest/azure"
@@ -100,16 +101,21 @@ func normalizeEnvironmentName(input string) string {
 }
 
 // AzureEnvironmentByName returns a specific Azure Environment from the specified endpoint
-func AzureEnvironmentByNameFromEndpoint(ctx context.Context, endpoint string, environmentName string) (*azure.Environment, error) {
+func AzureEnvironmentByNameFromEndpoint(ctx context.Context, metadataHost string, environmentName string) (*azure.Environment, error) {
 	if env, ok := sdkEnvironmentLookupMap[strings.ToLower(environmentName)]; ok {
 		return &env, nil
 	}
 
-	if endpoint == "" {
+	if metadataHost == "" {
 		return nil, fmt.Errorf("unable to locate metadata for environment %q from the built in `public`, `usgoverment`, `china` and no custom metadata host has been specified", environmentName)
 	}
 
-	environments, err := getSupportedEnvironments(ctx, endpoint)
+	metadataHostURL, err := url.Parse(metadataHost)
+	if err != nil {
+		return nil, err
+	}
+
+	environments, err := getSupportedEnvironments(ctx, *metadataHostURL)
 	if err != nil {
 		return nil, err
 	}
@@ -119,7 +125,7 @@ func AzureEnvironmentByNameFromEndpoint(ctx context.Context, endpoint string, en
 		if strings.EqualFold(env.Name, environmentName) || (environmentName == "" && len(environments) == 1) {
 			// if resourceManager endpoint is empty, assume it's the provided endpoint
 			if env.ResourceManager == "" {
-				env.ResourceManager = fmt.Sprintf("https://%s/", endpoint)
+				env.ResourceManager = metadataHostURL.String()
 			}
 
 			aEnv, err := buildAzureEnvironment(env)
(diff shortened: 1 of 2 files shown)


authentication/environment.go
```diff
@@ -100,16 +101,21 @@ func normalizeEnvironmentName(input string) string {
 }
 
 // AzureEnvironmentByName returns a specific Azure Environment from the specified endpoint
-func AzureEnvironmentByNameFromEndpoint(ctx context.Context, endpoint string, environmentName string) (*azure.Environment, error) {
+func AzureEnvironmentByNameFromEndpoint(ctx context.Context, metadataHost string, environmentName string) (*azure.Environment, error) {
```
tombuildsstuff, line 104: FYI the `authentication` package within this repository is being phased out in favour of the `auth` and `environments` packages in `hashicorp/go-azure-sdk` - where [the new `environments.FromEndpoint` method](https://github.com/hashicorp/go-azure-sdk/blob/6cdea3ca229b352fa45cf3efcc2ece9b0eaca92a/sdk/environments/from_endpoint.go#L17) actually [uses this approach](https://github.com/hashicorp/go-azure-sdk/blob/6cdea3ca229b352fa45cf3efcc2ece9b0eaca92a/sdk/internal/metadata/client.go#L22).

Since this new method is already threaded through `terraform-provider-azurerm`, I'm not sure that we need to make this change here - I think the only difference between the two is that `IsAzureStack` isn't yet implemented in `hashicorp/go-azure-sdk` - @manicminer did you have any thoughts on how we wanted to implement that?

magodo: @tombuildsstuff Yes, that is the reason why I bring this change here as the provider still uses the method here to detect whether it is an Azure Stack env.

manicminer: In terms of implementation, I was planning to update the helper func `IsEnvironmentAzureStack()` to use the new SDK package to retrieve the environment, and then remove the `AzureEnvironmentByNameFromEndpoint()`/`getSupportedEnvironments()` functions here - but it slipped through the cracks in the last couple weeks. I don't think there's any point updating the code here - I would suggest closing this PR and working on the `github.com/hashicorp/go-azure-sdk/sdk/environments` and `github.com/hashicorp/go-azure-sdk/sdk/internal/metadata` packages instead.

As for allowing the metadata host to be specified as a complete URI instead of just the hostname and assuming the scheme, path etc - I think that's probably fine. I thought about implementing it that way but decided initially to leave it since nobody really ever asked for this yet. Out of curiosity, what is the particular use-case here? You mention running a custom metadata service, but that sounds like something more likely to happen in Azure Stack rather than a proper cloud?

magodo: @manicminer Thank you for the context! That makes sense to close this PR and focus on the `github.com/hashicorp/go-azure-sdk`. Wondering if any existing issue that I can keep track?

Regarding the use case, we have an internal customer ask that they want to target to a regional ARM endpoint to workaround some issue (probably for the cross region metadata sync issue). Therefore I've just implemented a simple reverse proxy to allow them to spin up as a custom metadata host, where they can manipulate any field they want: https://github.com/magodo/azure-metadata-proxy

tombuildsstuff: I've [opened this PR in `go-azure-sdk`](https://github.com/hashicorp/go-azure-sdk/pull/355) which ports the `IsAzureStack` function over - at which point we should get this logic "for free" since the updated environments logic there supports sourcing this from a given endpoint.

With regards to changing the Metadata logic outside of Terraform to use a regional ARM endpoint, given how RP's and API Versions can be deployed only to specific Regions, it's likely that'd cause unexpected issues (which are likely to become a support burden/wild goose chase tracking these issues down), particularly in more complex configurations.

Whilst we believe it is possible to use the regional ARM endpoints in a targeted fashion, doing this for all resources is likely to both become a support burden (given that not every API version is deployed into every region, regional caching issues etc) - and as such we'd discourage users from this approach for now since we'd be unable to provide support for these issues.

Instead it likely makes more sense to use the ARM regional endpoint in a targeted fashion, presumably when listing the Key Vaults within a given Resource Group, based on the Location returned for the Resource Group, which I'm assuming is what this is trying to solve - and is something we should be able to do without changing the metadata endpoint?

magodo: The expected usage I can imagine is that users should sparingly use this feature, only when they are sure for a specific resource, the regional endpoint can resolve some Azure API issue, then they can use a dedicated provider alias for that certain resource.

magodo:
I've just opened a PR on top of @tombuildsstuff's change in the `go-azure-sdk`, to extend the `metadata_host` to be a URL: https://github.com/hashicorp/terraform-provider-azurerm/pull/20865

PR #155, opened 2023-03-02
resourcemanager/resourceids - better error messages during failure

tombuildsstuff:
This is still a WIP - but the intention is to provide better error messages during the parsing of Resource ID's into Segments - opening this as a placeholder for now


resourcemanager/resourceids/errors.go
```diff
@@ -0,0 +1,143 @@
...
+
+Expected a %[1]s ID that matched:
+
+> %[3]s
+
+However this value was provided:
+
+> %[4]s
+
+However the parsed Resource ID was missing a value for the %[2]q segment
```
mbfrahry, line 52: This reads weird with the `However` up above. I think we can just drop it
```suggestion
The parsed Resource ID was missing a value for the %[2]q segment
```

mbfrahry (dismissed):
I'm a fan of these changes but it did leave me a little more confused the way it is written.

I don't think we  should use `Segments` in the error as they are only used by the provider.

From your PR description, `However the parsed Resource ID was missing a value for the "staticMicrosoftServiceBus" segment (which should be the name of the Resource Provider [for example 'Microsoft.ServiceBus']).`, `staticMicrosoftServiceBus` is not used anywhere in the ID provided and threw me off trying to figure out where the issue was even though you stated it at the end of that description. 

Are we able to just point to the part of the ID that has an issue instead of pointing to the Segment that is wrong? Or we can surface what the segment chain should look like?

tombuildsstuff:
@mbfrahry 

> I don't think we should use Segments in the error as they are only used by the provider.

Segments is a pretty standard term for the components within a URI, so I think that's a reasonable assumption of familiarity here?

> From your PR description, However the parsed Resource ID was missing a value for the "staticMicrosoftServiceBus" segment (which should be the name of the Resource Provider [for example 'Microsoft.ServiceBus'])., staticMicrosoftServiceBus is not used anywhere in the ID provided and threw me off trying to figure out where the issue was even though you stated it at the end of that description.
>
> Are we able to just point to the part of the ID that has an issue instead of pointing to the Segment that is wrong? Or we can surface what the segment chain should look like?

Perhaps saying "the Nth segment should be a X (which should be a Y)" would work?

tombuildsstuff:
The latest changes improve things quite considerably when the incorrect Resource ID is specified:

Screenshot 2023-05-09 at 13 51 24

And also when the casing of a given segment is incorrect:

Screenshot 2023-05-09 at 14 04 10

When an extra segment is specified, this is now output as:

Screenshot 2023-05-09 at 14 05 27

tombuildsstuff:
The latest changes improve things quite considerably when the incorrect Resource ID is specified:

Screenshot 2023-05-09 at 13 51 24

And also when the casing of a given segment is incorrect:

Screenshot 2023-05-09 at 14 04 10

When an extra segment is specified, this is now output as:

Screenshot 2023-05-09 at 14 05 27

resourcemanager/resourceids/parse.go
```diff
@@ -238,7 +262,7 @@ func (p Parser) parseSegment(segment Segment, rawValue string, insensitively boo
 				return &*segment.FixedValue, nil
```
jackofallops, line 262: Not touched here, but I think this should be simplified to: 
```suggestion
				return segment.FixedValue, nil
```

jackofallops (approved):
Thanks @tombuildsstuff - just one comment on something I noticed in an unmodified line to possibly take a look at, otherwise LGTM 👍

mbfrahry (commented):
Re-reviewing just for some additional feedback post merge! 

I'm a fan of the errors being more clear on where the actual issue is andI think now we just need to clean up some of the fluff. The last image you shared posted the id three different times which just leads to a lot of clutter that isn't needed. 
image

 I imagine that process will involve cleanup in the actual resource code rather than anything to do here.

tombuildsstuff:
FWIW the initial stuttering (`parsing Resource ID %q: parsing %q:`) has been addressed within the Provider, I'd agree the extra segments could do with a better error message, however

PR #156, opened 2023-03-02
resourcemanager/commonids: adding Key Vault Key Version

tombuildsstuff:
Fixes https://github.com/hashicorp/pandora/pull/2183/files#r1122986709

### DIFF
--- a/resourcemanager/commonids/key_vault.go
+++ b/resourcemanager/commonids/key_vault.go
@@ -1,3 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
 package commonids
 
 import (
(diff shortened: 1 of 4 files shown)


resourcemanager/commonids/key_vault_test.go
```diff
@@ -6,9 +9,6 @@ import (
 	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
 )
 
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See NOTICE.txt in the project root for license information.
-
```
stephybun, line 11: Are these supposed to be removed?

stephybun (approved):
One question otherwise LGTM 🎍

resourcemanager/commonids/key_vault_test.go
tombuildsstuff: yep, they were supposed to be switched in #154 but I forgot to push the commit 🙃

PR #158, opened 2023-03-06
support resource providers cache

ms-henglu:
This PR moved the resource providers cache logic from `terraform-provider-azurerm`, so we can have the same approaches for both location and rp cache.

It also refactored the logic to ensure resource providers are registered, after moving from `terraform-provider-azurerm`, it can reuse the cached resource providers.

### DIFF
--- a/resourceproviders/cache.go
+++ b/resourceproviders/cache.go
@@ -1,0 +1,31 @@
+package resourceproviders
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/profiles/2017-03-09/resources/mgmt/resources"
+)
+
+// cachedResourceProviders can be (validly) nil - as such this shouldn't be relied on
+var cachedResourceProviders *[]resources.Provider
+
+// CacheSupportedProviders attempts to retrieve the supported Resource Providers from the Resource Manager API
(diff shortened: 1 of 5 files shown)


tombuildsstuff:
hey @ms-henglu 

Apologies for the delayed reply to this - it's taken a little time to review the PR's on `hashicorp/go-azure-helpers`.

As mentioned in https://github.com/hashicorp/terraform-provider-azurerm/pull/20050, having taken a look through this one it appears that moving this logic here would cause a circular reference when we come to migrate this logic across to `hashicorp/go-azure-sdk`, so unfortunately our previous suggestion was wrong here - apologies for that.

To fix this I've followed up your original PR with https://github.com/hashicorp/terraform-provider-azurerm/pull/21695 - which both adds the Resource Provider cache, migrates this logic across to using `hashicorp/go-azure-sdk`, updates the API Version being used to `2022-09-01` and fixes a bug during the Resource Provider registration - which'll fix this in a different manner.

As such whilst I'd like to thank you for this contribution, I'm going to close this PR in favour of https://github.com/hashicorp/terraform-provider-azurerm/pull/21695 which fixes this in a slightly different manner.

Thanks!

PR #161, opened 2023-03-14
AzureEnvironmentFromEndpoint

feiyushi:
- the latest API version of ARM metadata endpoint returns only one environment that the endpoint host belongs to.
- Added a new function to use the new API version so that it takes off the dependency on the cloud name parameter.

### DIFF
--- a/authentication/environment.go
+++ b/authentication/environment.go
@@ -134,6 +134,17 @@ func AzureEnvironmentByNameFromEndpoint(ctx context.Context, endpoint string, en
 	return nil, fmt.Errorf("unable to locate metadata for environment %q from custom metadata host %q", environmentName, endpoint)
 }
 
+// AzureEnvironmentFromEndpoint returns a specific Azure Environment from the specified endpoint
+// The difference between this and AzureEnvironmentByNameFromEndpoint is that
+// this function doesn't require an environment name as input
+func AzureEnvironmentFromEndpoint(ctx context.Context, endpoint string) (*azure.Environment, error) {
+	environment, err := getSupportedEnvironment(ctx, endpoint)
+	if err != nil {
+		return nil, err
+	}
+	return buildAzureEnvironment(environment)
+}
+
 // IsEnvironmentAzureStack returns whether a specific Azure Environment is an Azure Stack environment
 func IsEnvironmentAzureStack(ctx context.Context, endpoint string, environmentName string) (bool, error) {
 	if _, ok := sdkEnvironmentLookupMap[strings.ToLower(environmentName)]; ok {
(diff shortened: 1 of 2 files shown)


feiyushi:
Test passes if run only this one test. when run together with others in `make test` it fails with error `environment_test.go:99: Error getting Endpoint: retrieving environments from Azure MetaData service: Get "https://management.azure.com/metadata/endpoints?api-version=2022-09-01": dial tcp: lookup management.azure.com on 172.24.128.1:53: read udp 172.24.131.127:56569->172.24.128.1:53: i/o timeout`
not sure what causes it.

tombuildsstuff:
hey @feiyushi 

Thanks for this PR - we're actually in the middle of moving off of this logic [in favour of this logic within `hashicorp/go-azure-sdk`](https://github.com/hashicorp/go-azure-sdk/blob/8dae65113a1563dd209b5d718a50b7fd94b3f54d/sdk/environments/from_endpoint.go#L13-L43) - as such rather than implementing this here, this change would need to be made there instead.

That said, whilst removing the `environmentName` parameter isn't necessarily problematic, so that we can understand the context here, would you be able to elaborate on the problem you're looking to solve with this change?

Thanks!

feiyushi:
> 

Thanks for taking a look!
I was looking at the logic for fields `cloud_environment_name` and `metadata_host` in the packer-plugin-azure and found the reference here. doc for `metadata_host`: 
```
The Hostname of the Azure Metadata Service (for example management.azure.com), used to obtain the Cloud Environment when using a Custom Azure Environment. This can also be sourced from the ARM_METADATA_HOST Environment Variable. Note: CloudEnvironmentName must be set to the requested environment name in the list of available environments held in the metadata_host.
```
I was investigating something and found that metadata endpoint has released a new api version which returns only one cloud environment per endpoint. This makes more sense than the old one and removes the need for specifying a cloud name. I figured this could be useful for packer-plugin-azure if we can drop the correlation between these two fields, especially for clouds that haven't been made public or announced. More of a minor update than the old functionality.
I don't mind moving this logic to the other repo (you can take the essence of this PR and move it there if you want). Will packer-plugin-azure reference the [go-azure-sdk](https://github.com/hashicorp/go-azure-sdk) package?

manicminer:
Thanks for proposing this @feiyushi, as @tombuildsstuff mentioned we have moved this logic over to go-azure-sdk (see the actual metadata client implementation [here](https://github.com/hashicorp/go-azure-sdk/blob/8dae65113a1563dd209b5d718a50b7fd94b3f54d/sdk/internal/metadata/client.go)). As such we aren't looking to update the code here and in fact will be removing the `authentication` package from this module entirely in the near future.

You are correct that the metadata schema changed from essentially a list of environments to a single environment (it has actually changed back, see [these stubs](https://github.com/hashicorp/go-azure-sdk/blob/8dae65113a1563dd209b5d718a50b7fd94b3f54d/sdk/internal/metadata/client_test.go#L19-L24) for examples of earlier schema versions).

As such, whilst the environment name is no longer needed when configuring a custom environment, we do still use it in the Azure Terraform providers to initially select from a well-known environment such as the de-facto `public`, as well as `usgovernmentl4`, `usgovernmentl5` and `china`. If an environment name _and_ a custom metadata host are supplied, we simply ignore the environment name. Hope this helps!

feiyushi:
no problem! closing.

PR #172, opened 2023-07-25
`resourcemanager/commonids`: adding new Common IDs for App Service / App Service Environment / App Service Plan

tombuildsstuff:
This PR introduces new Common IDs for App Service, App Service Environment and App Service Plan.

This also introduces a couple of wrappers for both Function Apps and Web Apps (since these are just App Service instances today, but doing so allows us to differentiate these internally if needed). Since `hashicorp/pandora` only allows for a single instance of a Resource ID to match one Common ID, these are mostly for user-facing code - internally Pandora will output these as an App Service ID (at least for now).

### DIFF
--- a/resourcemanager/commonids/app_service_plan.go
+++ b/resourcemanager/commonids/app_service_plan.go
@@ -1,0 +1,127 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package commonids
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+)
+
+var _ resourceids.ResourceId = AppServicePlanId{}
+
+// AppServicePlanId is a struct representing the Resource ID for an App Service Plan
+type AppServicePlanId struct {
+	SubscriptionId    string
+	ResourceGroupName string
+	ServerFarmName    string
+}
+
+// NewAppServicePlanID returns a new AppServicePlanId struct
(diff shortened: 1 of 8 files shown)


resourcemanager/commonids/app_service_plan.go
```diff
@@ -0,0 +1,127 @@
...
+	if _, err := ParseAppServicePlanID(v); err != nil {
+		errors = append(errors, err)
+	}
+
+	return
+}
+
+// ID returns the formatted App Service Plan ID
+func (id AppServicePlanId) ID() string {
+	fmtString := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Web/serverFarms/%s"
```
jackofallops, line 101: Discussed offline at length, but calling out again here for completeness. Since `serverfarm` is considered "one word" in the service, this will need a state migration and careful testing as some older aspects of the API may be case sensitive.

jackofallops (approved):
Thanks @tombuildsstuff - Just the one comment, but LGTM 👍

resourcemanager/commonids/app_service_plan.go
tombuildsstuff: Yeah indeed - the Portal is using `serverFarms` at this point in time, which whilst I'm aware isn't a definitive source (and the Swagger is) - given the ARM conventions around resource naming (that we should be using camelCase) I think it's worth us running with that in this instance - but yeah, we _are_ going to need state migrations (and to parse the IDs insensitively in the Read functions etc)

tombuildsstuff: (chatting offline, this is fine)

PR #173, opened 2023-08-02
`resourcemanager/commonids`: Common IDs for Storage Account & Storage Container

tombuildsstuff:
This PR introduces 2 new Common IDs for Storage Account and Storage Container - whilst I've considered including File, Queue and Table into this list - those aren't really used outside the `storage` package, so there's less value in moving those here.


manicminer (approved):
LGTM 🥧

PR #174, opened 2023-08-04
Add support for optional/computed zone

mbfrahry:
There is an issue with `azurerm_mysql_flexible_server` where if you don't specify `zone`, a random zone gets applied which causes diffs on subsequent Terraform runs.

### DIFF
--- a/resourcemanager/commonschema/zone.go
+++ b/resourcemanager/commonschema/zone.go
@@ -8,8 +8,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
-// NOTE: we intentionally don't have an Optional & Computed here for behavioural consistency.
-
 // ZoneSingleRequired returns the schema used when a single Zone must be specified
 func ZoneSingleRequired() *schema.Schema {
 	return &schema.Schema{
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @mbfrahry 

We intentionally don't support Optional/Computed for Zones so that users have to make an explicit choice to ignore the Zone an instance is deployed into, to avoid unintentionally deploying (for example) a Primary and Secondary system into the same Azure Region.

In this case PostgreSQL Flexible Server will pick a Zone if one isn't specified, as such users can either opt to specify the Zone retroactively (to resolve the diff) or, or they can use `ignore_changes` to use the Azure provided zone - this [is called out in the documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/postgresql_flexible_server#zone) and an expected behaviour.

Since these fields are intentionally not Computed I'm going to close this PR for the moment - but this field is explicitly not marked as Computed to avoid unintentional downtime on a users behalf if a Zone becomes unavailable, so they need to make an explicit choice to either run with the Azure provided Zone (in which case they can use `ignore_changes` to accept the risk of deploying items into any Zone, rather than specifying the Zone explicitly), specifying the Zone retroactively (to resolve the diff) - or specifying the Zone explicitly from the start.

Thanks!

PR #176, opened 2023-09-08
`resourcemanager/commonids`: adding new Common ID for Virtual Machine Scale Set

mbfrahry:


tombuildsstuff (approved):
LGTM - pushed a commit to fix the `goimports`

PR #183, opened 2023-09-15
`resourcemanager/commonids`: adding Common IDs for Bot Service & Bot Service Channel

tombuildsstuff:


mbfrahry (changes requested):
Looks good but I think we should get the tests added for this too

stephybun (approved):

PR #186, opened 2023-10-16
New Common IDs for Kusto Cluster and Kusto Database

tombuildsstuff:
Reduces external package churn as seen in https://github.com/hashicorp/terraform-provider-azurerm/pull/23381


manicminer (approved):
LGTM 👍

PR #191, opened 2023-11-06
New Common ID: shared image gallery

jiaweitao001:


resourcemanager/commonids/shared_image_gallery.go
```diff
@@ -0,0 +1,140 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+)
+
+var _ resourceids.ResourceId = SharedImageGalleryId{}
+
+// SharedImageGalleryId is a struct representing the Resource ID for a Shared Image Gallery
+type SharedImageGalleryId struct {
```
tombuildsstuff, line 16: this is specific to a DevCenter, there's already a `SharedImageGallery` which is more widely used across the Provider, so this'll want to become:

```suggestion
type DevCenterSharedImageGalleryId struct {
```

Can we update this file and the filename to account for this?

jiaweitao001: > this is specific to a DevCenter, there's already a `SharedImageGallery` which is more widely used across the Provider, so this'll want to become:
> 
> Can we update this file and the filename to account for this?

Hi @tombuildsstuff , I just realized that it's not the correct resource id I actually would like to use. I'll correct it and update the PR accordingly.

jiaweitao001: Just updated the PR, the correct resource should be `azurerm_shared_image_gallery`.

tombuildsstuff (approved):
LGTM 👍

PR #204, opened 2024-01-17
commonids - add common IDs for chaos studio targets and capabilities

stephybun:
Needed for partial generation of capabilities and potentially experiments (if we generate it)


jackofallops (approved):
LGTM 🦬

PR #208, opened 2024-01-24
commonids: add composite resource id

catriona-m:


resourcemanager/commonids/composite_resource_id.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Parse the second of the two Resource IDs from the components
+	secondParser := resourceids.NewParserFromResourceIdType(output.Second)
+	secondParseResult, err := secondParser.Parse(components[1], insensitively)
+	if err != nil {
+		return nil, fmt.Errorf("parsing second id of CompositeResourceID: %v", err)
+	}
+	err = output.Second.FromParseResult(*secondParseResult)
+	if err != nil {
+		return nil, fmt.Errorf("populating second id of CompositeResourceID: %v", err)
```
tombuildsstuff, line 64: could we output the parsed id part as a part of the error to make this clearer?

```suggestion
		return nil, fmt.Errorf("parsing first id part %q of CompositeResourceID: %v", components[0], err)
	}
	err = output.First.FromParseResult(*firstParseResult)
	if err != nil {
		return nil, fmt.Errorf("populating first id part %q of CompositeResourceID: %v", components[0], err)
	}

	// Parse the second of the two Resource IDs from the components
	secondParser := resourceids.NewParserFromResourceIdType(output.Second)
	secondParseResult, err := secondParser.Parse(components[1], insensitively)
	if err != nil {
		return nil, fmt.Errorf("parsing second id part %q of CompositeResourceID: %v", components[1], err)
	}
	err = output.Second.FromParseResult(*secondParseResult)
	if err != nil {
		return nil, fmt.Errorf("populating second id part %q of CompositeResourceID: %v", components[1], err)
```

resourcemanager/commonids/composite_resource_id.go
```diff
@@ -0,0 +1,68 @@
...
+
+func ParseCompositeResourceID[T1 resourceids.ResourceId, T2 resourceids.ResourceId](input string, first T1, second T2) (*CompositeResourceID[T1, T2], error) {
+	return Parse(input, first, second, false)
+}
+
+func ParseCompositeResourceIDInsensitively[T1 resourceids.ResourceId, T2 resourceids.ResourceId](input string, first T1, second T2) (*CompositeResourceID[T1, T2], error) {
+	return Parse(input, first, second, true)
+}
+
+func Parse[T1 resourceids.ResourceId, T2 resourceids.ResourceId](input string, first T1, second T2, insensitively bool) (*CompositeResourceID[T1, T2], error) {
```
tombuildsstuff, line 33: given we're providing wrapper functions above, does this method need to be public?

resourcemanager/commonids/composite_resource_id.go
```diff
@@ -0,0 +1,68 @@
...
+func (id CompositeResourceID[T1, T2]) String() string {
+	fmtString := "%s\n%s"
+	return fmt.Sprintf(fmtString, id.First.String(), id.Second.String())
+}
+
+func ParseCompositeResourceID[T1 resourceids.ResourceId, T2 resourceids.ResourceId](input string, first T1, second T2) (*CompositeResourceID[T1, T2], error) {
+	return Parse(input, first, second, false)
+}
+
+func ParseCompositeResourceIDInsensitively[T1 resourceids.ResourceId, T2 resourceids.ResourceId](input string, first T1, second T2) (*CompositeResourceID[T1, T2], error) {
```
tombuildsstuff, line 29: could we document this struct/the functions to explain how this is intended to be used?

tombuildsstuff (commented):
hey @catriona-m 

Thanks for this PR - apologies for the delayed review here.

Taking a look through on the whole this is looking pretty good, if we can fix up the comments then this should otherwise be good to go 👍 

Thanks!

resourcemanager/commonids/composite_resource_id.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+)
+
+// CompositeResourceID is a struct representing the Resource ID for a Composite Resource Id
+type CompositeResourceID[T1 resourceids.ResourceId, T2 resourceids.ResourceId] struct {
+	First  T1
+	Second T2
```
tombuildsstuff, line 13: _minor_ but could we document these fields so that it's clearer what these are?

```suggestion
	// First specifies the first component of this Resource ID.
	// This is in the format `{first}|{second}`.
	First  T1
	
	// Second specifies the second component of this Resource ID
	// This is in the format `{first}|{second}`.
	Second T2
```

tombuildsstuff (approved):
Left a few minor comments inline but otherwise 👍

resourcemanager/commonids/composite_resource_id.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+)
+
+// CompositeResourceID is a struct representing the Resource ID for a Composite Resource Id
+type CompositeResourceID[T1 resourceids.ResourceId, T2 resourceids.ResourceId] struct {
+	// First specifies the first component of this Resource ID.
+	// This is in the format `{first}|{second}`.
+	First T1
+	// Second specifies the second component of this Resource ID
```
tombuildsstuff, line 15: very pedantic but:

```suggestion

	// Second specifies the second component of this Resource ID
```

tombuildsstuff (approved):
One really minor nit for readability purposes but this otherwise LGTM 👍

PR #213, opened 2024-02-06
commonids: add commonid for BillingAccountCustomerId

catriona-m:


tombuildsstuff (approved):
LGTM - @manicminer mind giving this one a second review from a naming perspective, given the number of different types of billing account ids?

manicminer (approved):
I think this is fine, we should probably have a different ID for different account types (e.g. enrollment accounts). LGTM 👍

PR #216, opened 2024-02-07
commonids: add commonid for `TenantId`

catriona-m:


tombuildsstuff:
@manicminer given the default Tenant ID can be represented as `/` - do we also want a `DefaultTenantID` perhaps?

manicminer (approved):
Agreed, for convenience it might be worth declaring a default tenant for easy re-use. Otherwise LGTM 👍 

```go
var DefaultTenantId = TenantId{
	TenantId: "/",
}
```

resourcemanager/commonids/tenant.go
```diff
@@ -0,0 +1,110 @@
...
+
+// TenantId is a struct representing the Resource ID for a Tenant
+type TenantId struct {
+	TenantId string
+}
+
+// DefaultTenantId returns a default TenantId
+var DefaultTenantId = TenantId{
+	TenantId: "/",
+}
```
tombuildsstuff, line 23: pedantic request: but rather than defining the fields inline can we use `NewTenantID` here? It ensures that were we to add another field here in the future that we'd get a compiler error, so this'd be updated?

```suggestion
var DefaultTenantId = NewTenantID("/")
```

catriona-m: makes sense 👍  I've updated that now

resourcemanager/commonids/tenant.go
```diff
@@ -0,0 +1,108 @@
...
+
+// DefaultTenantId returns a default TenantId
+var DefaultTenantId = NewTenantID("/")
+
+// NewTenantID returns a new TenantId struct
+func NewTenantID(tenantId string) TenantId {
+	return TenantId{
+		TenantId: tenantId,
+	}
+}
```
tombuildsstuff, line 28: Thinking about the default Tenant ID perhaps we want to handle this differently? If we make the `TenantId` field a pointer, we could support parsing both the Default Tenant ID and another Tenant ID? It'd be different to the other Resource ID parsers but this is a unique case where the Default is handled differently, which I don't believe we've got anywhere else - so perhaps it's worth handling that explicitly here?

```suggestion
// TenantId is a struct representing the Resource ID for a Tenant
type TenantId struct {
	TenantId *string
}

// NewDefaultTenantId returns a TenantId for the default Tenant
func NewDefaultTenantID() TenantId {
	return TenantId{}
}

// NewTenantID returns a new TenantId struct
func NewTenantID(tenantId string) TenantId {
	return TenantId{
		TenantId: &tenantId,
	}
}
```

WDYT @catriona-m @manicminer?

Issue #189, opened 2023-11-03
 Dynamic Resource IDs / a replacement for `azure.ValidateResourceID`


tombuildsstuff:
The new Resource ID parsing logic defines Go Structs for each Resource ID Type, which allow parsing a Resource ID based on a known set of Resource ID segments, allowing us to provide differentiation based on the Resource ID Segment Type in the future.

However some Azure Resources support specifying _any_ Azure Resource ID as an argument, meaning that today we have a validation function in the form of `azure.ValidateResourceID` which is used to ensure that the argument looks _vaguely_ like an Azure Resource ID, but only ensures that we have a matching number of Key/Value Pairs.

Resource ID Scopes follow a similar pattern, with the Scope being an Azure Resource ID of some description - which (at this point in time) gets parsed out as a literal string value - but isn't validated. The behaviour of both of these differs from the rest of the new Resource ID parsing logic, which implies a known set of Resource ID Segments for a given Resource ID - since these can be any Azure Resource ID.

This means that we're currently unable to use the `ResourceIDReference` Common Schema types and that we're unable to offer dynamic recasing to workaround issues where the Azure API returns a different casing than is defined in the Terraform configuration.

---

[This commit](https://github.com/hashicorp/go-azure-helpers/commit/87a85ae3b211d70526694d9bfcd0834fe4201635) introduces a proof-of-concept dynamic recaser, which means that we can dynamically recase a Resource ID, for example:

> recaser.ReCase("/subscriptions/11111/resourcegroups/bobby/providers/Microsoft.Compute/availabilitySets/HeYO")

gets output as:

> /subscriptions/11111/resourceGroups/bobby/providers/Microsoft.Compute/availabilitySets/HeYO

This is achieved by having an `init()` in each Resource ID type register itself with the recaser, meaning that as Resource ID Types are imported, these get registered, meaning those Resource IDs can be used in this recasing function - which can be used on scopes, fixing issues such as https://github.com/hashicorp/terraform-provider-azurerm/pull/23644.

This logic should also be reusable for an updated `ValidateGenericAzureResourceID` function too - which presumably requires parsing the Resource ID twice, once case-insensitively (to check that the Resource ID type is valid, where an error can mean it's the wrong type) - and a second time to ensure the casing matches what we expect (which can be surfaced as an error).

[...]

PR #221, opened 2024-03-14
Add support for dynamic resource id recasing

catriona-m:
partially addresses #189


resourcemanager/recaser/recaser_test.go
```diff
@@ -0,0 +1,129 @@
+package recaser_test
+
+import (
+	"testing"
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/recaser"
+)
+
+func TestRecaserWithIncorrectCasing(t *testing.T) {
```
tombuildsstuff, line 9: could we add a couple of the awkward test cases:

1. There's a handful of APIs that return the full URI with a port in some cases e.g. `https://management.azure.com:80/subscriptions/12345`
2. Could we add a Data Plane URI here to ensure this is ignored too: `https://example.blob.storage.azure.com/container1` should be fine

As such could we add a couple of test cases covering those?

resourcemanager/recaser/registration.go
```diff
@@ -0,0 +1,36 @@
...
+func init() {
+	//register common ids
+	resourceIdsWriteLock.Lock()
+	for _, id := range commonids.CommonIds() {
+		key := strings.ToLower(id.ID())
+		if _, ok := knownResourceIds[key]; !ok {
+			knownResourceIds[key] = id
+		}
+	}
+	resourceIdsWriteLock.Unlock()
```
tombuildsstuff, line 24: perhaps pedantic, but is there a reason not to call `RegisterResourceId` here instead?

```suggestion
	for _, id := range commonids.CommonIds() {
		RegisterResourceId(id)
	}
```

resourcemanager/recaser/recase.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+)
+
+// ReCase tries to determine the type of Resource ID defined in `input` to be able to re-case it from
+func ReCase(input string) string {
+	return ReCaseWithIds(input, knownResourceIds)
+}
+
+// ReCaseWithIds tries to determine the type of Resource ID defined in `input` to be able to re-case it from based on an input list of Resource IDs
+func ReCaseWithIds(input string, ids map[string]resourceids.ResourceId) string {
```
tombuildsstuff, line 17: Can we make this function private/unexported? This should only be used for testing purposes, else we might end up using this unintentionally (which would be a bug, since it'd mean we're not registering all of the Resource IDs)

```suggestion
// reCaseWithIds tries to determine the type of Resource ID defined in `input` to be able to re-case it from based on an input list of Resource IDs
func reCaseWithIds(input string, ids map[string]resourceids.ResourceId) string {
```

resourcemanager/recaser/registration.go
catriona-m: makes sense 👍 fixed now!

resourcemanager/recaser/recase.go
```diff
@@ -0,0 +1,101 @@
...
+	if strings.Contains(strings.ToLower(input), strings.ToLower(segment)) {
+		re := regexp.MustCompile(fmt.Sprintf("(?i)%s", segment))
+		input = re.ReplaceAllString(input, segment)
+	}
+	return input
+}
+
+// buildInputKey takes an input id string and removes user-specified values from it
+// so it can be used as a key to extract the correct id from knownResourceIds
+func buildInputKey(input string) string {
```
tombuildsstuff, line 79: perhaps we should update this to also return a boolean for whether this is usable or not?

```suggestion
func buildInputKey(input string) (*string, bool) {
```

That'd allow this to become:

```
key, ok := buildInputKey(input)
if ok {
  // assuming that `key` had a value so *key should be safe
}

File: resourcemanager/recaser/recase.go [hunk 5]
```diff
@@ -0,0 +1,101 @@
...
+
+	segments := strings.Split(input, "/")
+	if len(segments)%2 != 0 {
+		for i := 1; len(segments) > i; i++ {
+			if i%2 != 0 {
+				key := segments[i]
+				output = fmt.Sprintf("%s/%s/", output, key)
+				if strings.EqualFold(key, "providers") && len(segments) >= i+2 {
+					value := segments[i+1]
+					output = fmt.Sprintf("%s%s", output, value)
```
tombuildsstuff, line 95: perhaps it's worth being consistent here?

```suggestion
					output = fmt.Sprintf("%s/%s", output, value)
```

resourcemanager/recaser/recase.go
```diff
@@ -0,0 +1,101 @@
...
+		return output
+	}
+
+	segments := strings.Split(input, "/")
+	if len(segments)%2 != 0 {
+		for i := 1; len(segments) > i; i++ {
+			if i%2 != 0 {
+				key := segments[i]
+				output = fmt.Sprintf("%s/%s/", output, key)
+				if strings.EqualFold(key, "providers") && len(segments) >= i+2 {
```
tombuildsstuff, line 93: might be worth a comment explaining the why here?

tombuildsstuff (approved):
A few minor points but this otherwise LGTM 👍

resourcemanager/recaser/recase.go
catriona-m: in this case we don't need to add `/` since there is no user input between `providers` and the next segment. I've added some comments to make the expected inputs and outputs clearer here

resourcemanager/recaser/registration.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"strings"
+	"sync"
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/commonids"
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+)
+
+var KnownResourceIds = make(map[string]resourceids.ResourceId)
```
tombuildsstuff, line 11: Out of interest, what's the reason for having this public/exported?

catriona-m: the compiler was giving an error, but you're right it looks like it doesn't really need to be public - fixed

resourcemanager/recaser/recase.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+)
+
+// ReCase tries to determine the type of Resource ID defined in `input` to be able to re-case it from
+func ReCase(input string) string {
+	return reCaseWithIds(input, knownResourceIds)
+}
+
+// reCaseWithIds tries to determine the type of Resource ID defined in `input` to be able to re-case it from based on an input list of Resource IDs
```
stephybun, line 16:
```suggestion
// reCaseWithIds tries to determine the type of Resource ID defined in `input` to be able to re-case it based on an input list of Resource IDs
```

stephybun (approved):
Thanks @catriona-m. One minor suggestion on the function comment but otherwise LGTM 👍

tombuildsstuff (approved):
LGTM 👍

PR #222, opened 2024-03-21
`resourceid`: parser supports data plane host segment

wuxu92:
This PR is to support parsing data plane resource IDs like keyvault/managed hsm keys, secrets for new resource ID format.

Required by: https://github.com/hashicorp/terraform-provider-azurerm/pull/25324


katbyte:
I think this can be closed now that https://github.com/hashicorp/terraform-provider-azurerm/pull/25601 has been merged?

PR #234, opened 2024-05-06
`resourcemanager/resourceids`: adding a `Match` function

tombuildsstuff:
The `Match` function takes two instances of the ResourceId type and allows comparing the parsed segments for these.

Whilst this is also possible by comparing the value of `.ID()` in both instances - that's not case-aware and so will be problematic in the future - hence the need for this function which can account for case-aware comparisons.

This also exposed a feature-flag for treating User Specified Segments as case-insensitive for comparison purposes - however this isn't usable at this time and so **must not** be exposed as a toggle at this point-in-time, since it'll cause more problems than it solves. However having this internal-only feature flag allows us to built out the surrounding components ahead-of-time and add tests covering both scenarios - therefore this is beneficial to have as an internal toggle today - even if it's not usable for end-users for some time.

There are currently 6 instances of this across the Provider:

Screenshot 2024-05-06 at 13 02 05

in addition to (at least) https://github.com/hashicorp/terraform-provider-azurerm/pull/25809 - therefore we should get ahead of this and switch to using `resourceids.Match` to enable us to handle this consistently in the future.

### DIFF
--- a/resourcemanager/features/user_specified_segments.go
+++ b/resourcemanager/features/user_specified_segments.go
@@ -7,7 +7,7 @@ package features
 // Resource ID Segments should be compared case-insensitively as required.
 //
 // @tombuildsstuff: whilst this IS EXPOSED in the public interface - this is NOT READY FOR USE and should
-// not be exposed in user-focused logic at this time, else this'll become a source of knock-on problems
+// not be exposed to users (i.e. as a feature-toggle) until this work is completed - as this'll become a source of knock-on problems
 // rather than being useful.
 //
 // There are a number of dependencies to enabling this, including completing the standardiation on the
--- a/resourcemanager/features/user_specified_segments.go
+++ b/resourcemanager/features/user_specified_segments.go
@@ -1,0 +1,15 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package features
+
+// TreatUserSpecifiedSegmentsAsCaseInsensitive is a feature-toggle which specifies whether User Specified
+// Resource ID Segments should be compared case-insensitively as required.
+//
+// @tombuildsstuff: whilst this IS EXPOSED in the public interface - this is NOT READY FOR USE and should
+// not be exposed in user-focused logic at this time, else this'll become a source of knock-on problems
+// rather than being useful.
+//
+// There are a number of dependencies to enabling this, including completing the standardiation on the
+// `ResourceId` interface and the `ResourceIDReference` schema types - and surrounding updates.
+var TreatUserSpecifiedSegmentsAsCaseInsensitive = false
--- a/resourcemanager/resourceids/match.go
+++ b/resourcemanager/resourceids/match.go
@@ -1,0 +1,61 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package resourceids
+
+import (
+	"reflect"
+	"strings"
+
+	"github.com/hashicorp/go-azure-helpers/resourcemanager/features"
+)
+
+// Match compares two instances of the same ResourceId and determines whether they are a match
+//
+// Whilst it might seem fine to compare the result of the `.ID()` function, that doesn't account
+// for Resource ID Segments which need to be compared as case-insensitive.
+//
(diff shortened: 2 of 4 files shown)


resourcemanager/features/user_specified_segments.go
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package features
+
+// TreatUserSpecifiedSegmentsAsCaseInsensitive is a feature-toggle which specifies whether User Specified
+// Resource ID Segments should be compared case-insensitively as required.
+//
+// @tombuildsstuff: whilst this IS EXPOSED in the public interface - this is NOT READY FOR USE and should
+// not be exposed in user-focused logic at this time, else this'll become a source of knock-on problems
```
stephybun, line 10: Could you explain what you mean by user-focused logic here?

resourcemanager/features/user_specified_segments.go
```diff
@@ -0,0 +1,15 @@
...
+// rather than being useful.
+//
+// There are a number of dependencies to enabling this, including completing the standardiation on the
+// `ResourceId` interface and the `ResourceIDReference` schema types - and surrounding updates.
+var TreatUserSpecifiedSegmentsAsCaseInsensitive = false
```
stephybun, line 15: Despite the comment I could still see this creeping into places where it shouldn't in the provider. Would it make sense and/or be possible for this to be accompanied by a check in the provider to flag new additions of this?

stephybun (approved):
Thanks @tombuildsstuff I have two questions around the case insensitive flag, otherwise LGTM 👍

resourcemanager/features/user_specified_segments.go
tombuildsstuff: By this, do you mean a check in the Provider to ensure that we're not doing `.ID() ==`? If so, yes - that can be added at the same time as fixing the 6 instances above

stephybun: I actually mean a check in the Provider to ensure we're not setting `features.TreatUserSpecifiedSegmentsAsCaseInsensitive = true` unless it's been explicitly added to a list of exceptions, like we do with [these checks](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/provider/provider.go)

resourcemanager/features/user_specified_segments.go
tombuildsstuff: Here I'm using the phrase `user-focused logic` to mean "surfaced to end-users" - for example exposing a feature-toggle to do so in the Provider, or setting this to true/doing this by default.

Perhaps this'd be better rephrased as:

```suggestion
// not be exposed to users (i.e. as a feature-toggle) until this work is completed - as this'll become a source of knock-on problems
```

WDYT?

---

An aside: initially I was trying to make the public surface for this a function (e.g. `features.TreatUserSpecifiedSegmentsAsCaseInsensitive()`, which in-turn set a feature-flag within an `internal` package in this project) - to ensure that we don't unintentionally use this toggle outside of this project - but I don't _think_ having this publicly accessible is the end of the world at this point-in-time.

stephybun: I agree, the rephrased comment makes the intention clearer 👍

resourcemanager/features/user_specified_segments.go
tombuildsstuff: Yep, makes sense - we probably want both checks actually.

It's worth noting that since this feature-flag is for the behaviour as a whole (i.e. across the entire Provider) rather than being for individual Resource ID types - we'd only set this one-time rather than in multiple places anyway.

resourcemanager/features/user_specified_segments.go
tombuildsstuff: 👍 committed this suggestion then

resourcemanager/features/user_specified_segments.go
tombuildsstuff: For info: I've intentionally not included the validation changes in https://github.com/hashicorp/terraform-provider-azurerm/pull/25873 since this wants to be done in a follow up PR - will send that in the morning:

PR #236, opened 2024-05-08
`commonids` - add `NewCompositeResourceID` function to construct composite resource IDs

stephybun:


resourcemanager/commonids/composite_resource_id_test.go
```diff
@@ -5,9 +5,41 @@ package commonids
...
+		t.Fatalf("expected First ID string to be %q but got %q", botIdString, id.First.ID())
+	}
+
+	if !strings.EqualFold(id.Second.ID(), appIdString) {
+		t.Fatalf("expected Second ID string to be %q but got %q", appIdString, id.Second.ID())
+	}
+
+	expectedIdString := "/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/example-resource-group/providers/Microsoft.BotService/botServices/botServiceValue|/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/example-resource-group/providers/Microsoft.Web/sites/siteValue"
+
+	if !strings.EqualFold(id.ID(), expectedIdString) {
```
tombuildsstuff, line 38: this should exact match?

```suggestion
	if id.ID != expectedIdString {
```

tombuildsstuff (approved):
One thing but this otherwise LGTM 👍

resourcemanager/commonids/composite_resource_id_test.go
stephybun: you're right, sorry habit

PR #238, opened 2024-05-24
`commonschema`: adding a new Common Schema type and Constant for Public Network Access

tombuildsstuff:
This allows us to consolidate onto this in time - and means that we have a single set of constants for this purpose


catriona-m (approved):
LGTM 👍

PR #246, opened 2025-02-06
`pointer` - add generics enum pointer helpers

jackofallops:


lang/pointer/generic.go
```diff
@@ -13,7 +13,31 @@ func From[T any](input *T) (output T) {
...
 // To is a generic function that returns a pointer to the value provided.
 func To[T any](input T) *T {
 	return &input
 }
+
+// ToEnum is a helper function to cast strings as an Enum type where API objects expect a pointer to the Enum value
+// example code simplification:
+// APIModel.SomeValue = pointer.To(someservice.SomeEnumType(d.Get["some_variable"].(string)))
+// becomes
+// APIModel.SomeValue = pointer.ToEnum[someservice.SomeEnumType](d.Get["some_variable"].(string))
```
stephybun, line 39: Should `d.Get["some_variable"].(string)` be `d.Get("some_variable").(string)`?
And should we be using syntax for typed resources over untyped ones
```suggestion
// APIModel.SomeValue = pointer.To(someservice.SomeEnumType(model.SomeVariable))
// becomes
// APIModel.SomeValue = pointer.ToEnum[someservice.SomeEnumType](model.SomeVariable)
```

jackofallops: It's just a godoc example, but examples for typed instead of legacy is probably a good shout. 👍

stephybun (approved):
Thanks @jackofallops LGTM 🍕

PR #251, opened 2025-03-21
Fix filepath

wyattfry:
Looks like a copy-paste error


jackofallops:
Thanks @wyattfry - Given this has never worked, and suddenly introducing (albeit fixing) an auto-updater process might cause some unexpected issues, I'm going to close this for now (and disable the workflow). We can discuss the value of bringing it back in with the team (since we've clearly not needed / used it since inception). We can take a look at the process as a whole after that, and re-open this as needed.

PR #275, opened 2026-03-17
Identity: implement the (Un)Marshaler on value type instead of pointer type

magodo:
## Community Note

## Description

Per https://go.dev/doc/effective_go#pointers_vs_values:

> The rule about pointers vs. values for receivers is that value methods can be invoked on pointers and values, but pointer methods can only be invoked on pointers.
> ...There is a handy exception, though. When the value is addressable, the language takes care of the common case of invoking a pointer method on a value by inserting the address operator automatically.

In the `json.Marshal()`, it will check each field's type whether it implements the `json.Marshler`. The pointer of the identity types implement it, while the value types are not, which causes the member identity field to fall back to the default marshal logic.

This issue was detected when implementing a resource (Storage Actions Task), where its Swagger declares the `identity` as required. The generated go-azure-sdk then defines the `identity` field as an identity type, instead of an identity pointer type.

This PR changes the receiver from pointer to value type. Together with updated test cases, which:

- Removes the nil input test case, which will always be marshaled into `null` by json.Marshal
- Changes the call to `json.Marshal` instead of direct call to MarshalJSON to make it more of a real use case
- Add a new test step that the `input` is a value type to cover this new issue
- Add an `expect` verification facet (arguably we shall remove the other verification facets as some tests will involve a customized Unmarshal).

## Change Log

Below please provide what should go into the changelog (if anything) conforming to the [Changelog Format documented here](../blob/main/contributing/topics/maintainer-changelog.md).

* added new function `AFunction` [GH-00000]

- [x] Bug Fix
- [ ] New Feature
- [ ] Enhancement
- [ ] Breaking Change

## Related Issue(s)
Fixes #0000

## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the provider.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

> [!NOTE] 
> If this PR changes meaningfully during the course of review please update the title and description as required.

### DIFF
--- a/resourcemanager/identity/legacy_system_and_user_assigned_list_test.go
+++ b/resourcemanager/identity/legacy_system_and_user_assigned_list_test.go
@@ -12,24 +12,28 @@ import (
 
 func TestLegacySystemUserAssignedListMarshal(t *testing.T) {
 	testData := []struct {
-		input                           *LegacySystemAndUserAssignedList
+		input                           any
+		expect                          map[string]any
 		expectedIdentityType            string
 		expectedUserAssignedIdentityIds []string
 		expectError                     bool
 	}{
 		{
-			input:                           nil,
-			expectedIdentityType:            "None",
-			expectedUserAssignedIdentityIds: []string{},
-		},
-		{
-			input:                           &LegacySystemAndUserAssignedList{},
+			input: &LegacySystemAndUserAssignedList{},
+			expect: map[string]any{
+				"type":                   "None",
+				"userAssignedIdentities": nil,
+			},
 			expectedIdentityType:            "None",
 			expectedUserAssignedIdentityIds: []string{},
 		},
 		{
 			input: &LegacySystemAndUserAssignedList{
 				Type: TypeNone,
+			},
+			expect: map[string]any{
+				"type":                   "None",
+				"userAssignedIdentities": nil,
 			},
 			expectedIdentityType:            "None",
 			expectedUserAssignedIdentityIds: []string{},
@@ -41,6 +45,10 @@ func TestLegacySystemUserAssignedListMarshal(t *testing.T) {
 					"first",
 				},
 			},
+			expect: map[string]any{
+				"type":                   "None",
+				"userAssignedIdentities": nil,
+			},
 			expectedIdentityType:            "None",
 			expectedUserAssignedIdentityIds: []string{
 				// intentionally empty since this is bad data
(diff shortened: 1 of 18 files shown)


wuxu92 (approved):
LGTM 👍

ms-henglu (approved):
Thanks for the PR! LGTM

wuxu92:
Although this PR looks good, many resources depend on these methods, so applying it to the entire provider poses a significant risk. @magodo Would it be better to add a workaround in Pandora that changes the value typed as `identity` to a pointer type for the Storage Actions Task resource? I think most of the identity blocks in the azure-rest-api-spec are optional and pointer type in SDK.

magodo:
Could you elaborate on: *applying it to the entire provider poses a significant risk*?

wuxu92:
> Could you elaborate on: _applying it to the entire provider poses a significant risk_?

I misspoke: it's a potential risk, not a significant one.

resourcemanager/identity/legacy_system_and_user_assigned_list_test.go
```diff
@@ -119,7 +166,7 @@ func TestLegacySystemUserAssignedListMarshal(t *testing.T) {
 	for i, v := range testData {
 		t.Logf("step %d..", i)
 
-		encoded, err := v.input.MarshalJSON()
+		encoded, err := json.Marshal(v.input)
```
sreallymatt, line 169: Given we have direct access here to the `MarshalJSON` implementation, could you add an additional call for that as well as a sanity check?

e.g. `encoded, err := json.Marshal ...` and `encodedDirect, err := v.input.MarshalJSON ...`

While this should be fine regardless, it preserves existing test logic which may not be a bad thing given there is no context as to why this was implemented this way.

(same for all tests)

sreallymatt (changes requested):

resourcemanager/identity/legacy_system_and_user_assigned_list_test.go
magodo: Good point! I've submit a new commit to add it.

sreallymatt (approved):
Thanks @magodo - I had similar reservations as Xu mentioned above, but I replaced `go-azure-helpers` in the provider with these changes locally and ran all tests suffixed with `_identity*` to double check these changes don't affect any existing `identity` blocks.

LGTM ✅

PR #279, opened 2026-06-09
keyvault: support for non-default ports

tombuildsstuff:
## Community Note

## Description

This commit fixes an issue where a non-default is being used but the port gets stripped.

This handles this in a backwards-compatible way by stripping the port from the response to normalise it when the port being used matches the default for the scheme being used.

For example, this means that `https://example:443` continues being flattened down to `https://example` - but `https://example:444` will be retained as-is.

## Change Log

Below please provide what should go into the changelog (if anything) conforming to the [Changelog Format documented here](../blob/main/contributing/topics/maintainer-changelog.md).

* `keyvault`: support for non-default ports

- [x] Bug Fix
- [x] Enhancement


jackofallops (approved):
Thanks @tombuildsstuff - LGTM 👍

PR #280, opened 2026-06-23
resourcemanager/tags, framework/commonschema: support provider-level ignore_tags

cdobbyn:
## Community Note

## Description

Adds support for a provider-level "ignore tags" capability to the shared tag helpers, mirroring the AWS provider's [ignoring changes in all resources](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#ignoring-changes-in-all-resources) feature.

This introduces a new `IgnoreConfig` in `resourcemanager/tags`:

* `IgnoreConfig{ Keys, KeyPrefixes }` — case-sensitive exact-key and `key_prefix` ("starts with") matching.
* A package-level `SetIgnore(*IgnoreConfig)` / `Ignore() *IgnoreConfig` hook so a provider can configure the ignore set once at provider-configure time.
* `ApplyMap` / `ApplyPtrMap` helpers that scrub ignored keys from a tag map (nil-safe on a nil receiver and nil map).

The existing tag `Expand`/`Flatten` helpers in both `resourcemanager/tags` and `framework/commonschema` now consult the configured `IgnoreConfig`:

* **Flatten (read):** ignored keys are scrubbed before being written to state, so an externally-applied tag never lands in state and never shows as drift.
* **Expand (create/update):** ignored keys are excluded from the desired set.

Applied symmetrically, an ignored tag is never added and is never reported as drift. When no config is set (the default), every code path is a no-op, so existing behaviour is unchanged.

This is the upstream half required by hashicorp/terraform-provider-azurerm#32664, which adds a provider-level `ignore_tags` block consuming this hook. That PR currently `replace`s this module with this branch temporarily; once this is merged and released, the `replace` will be dropped and the version bumped.

Unit tests are included for `IgnoreConfig` (exact keys, key prefixes, union, case-sensitivity, nil/empty no-op, nil map) and for `Expand`/`Flatten` honouring an active config, on both the `resourcemanager/tags` and `framework/commonschema` helpers.

## Change Log

* `resourcemanager/tags`: support for provider-level ignoring of tag keys and key prefixes via `IgnoreConfig` and the package-level `SetIgnore`/`Ignore` hook [GH-00000]
* `framework/commonschema`: tag `Expand`/`Flatten` now honour the configured `IgnoreConfig` [GH-00000]

- [ ] Bug Fix
- [x] New Feature
- [ ] Enhancement
- [ ] Breaking Change

## Related Issue(s)

Backs hashicorp/terraform-provider-azurerm#32664 (and the feature request hashicorp/terraform-provider-azurerm#28711).

## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the provider.

## Changes to Security Controls

No changes to security controls. The change only filters which tag keys are written to / read from Terraform state based on user-supplied configuration; no access controls, encryption, or logging are affected.

### DIFF
--- a/framework/commonschema/tags.go
+++ b/framework/commonschema/tags.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/hashicorp/go-azure-helpers/framework/convert"
 	"github.com/hashicorp/go-azure-helpers/framework/typehelpers"
+	rmtags "github.com/hashicorp/go-azure-helpers/resourcemanager/tags"
 	"github.com/hashicorp/terraform-plugin-framework-validators/mapvalidator"
 	datasourceschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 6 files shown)


cdobbyn:
@catriona-m would you be able to look at this or suggest someone else? This addresses an issue we've seen from a number of our azure customers where they want to have specific managed tags that we simply need to ignore from our terraform state. It's related to a PR on the actual azurerm provider. Any help would be much appreciated. Thank you.

PR #283, opened 2026-07-06
`dependencies` update go version to `1.26.4` and terraform-plugin-* to align with azurerm

jackofallops:


catriona-m (approved):
Thanks @jackofallops LGTM!

